PR #24, opened 2026-04-22
SDD: long form Bluesky strategy (exploration spec)

kraftbj:
Fixes DOTCOM-16885
See DOTCOM-16810 (parent epic)

## Summary

Exploration SDD for [DOTCOM-16810](https://linear.app/a8c/issue/DOTCOM-16810) — how FOSSE should publish WordPress posts that don't fit Bluesky's ~300-grapheme limit.

This is the long-form half of the [Bluesky-native-publishing epic](https://linear.app/a8c/issue/DOTCOM-16795). Short-form shipped in #18/#21; long-form was punted to this follow-up.

**Opened with open questions and a stated lean, then revised after the [2026-04-23 Jim Ray call](https://fossep2.wordpress.com/2026/04/23/call-notes-bluesky-intro-jim-ray/).** That call clarified that Option 3's enabling renderer is months out and multi-iteration — we're not bridging a short gap, we're bridging a long one. The spec now recommends Option 5 (teaser mini-thread) as the v1 default.

## What's in this PR

- `sdd/long-form-bluesky-strategy/requirements.md` — gathered requirements, assumptions, open questions.
- `sdd/long-form-bluesky-strategy/spec.md` — all 5 candidate options analyzed (pros / cons / engineering cost / upstream work), recommended v1 default + reasoning, resolved open questions, and explicit holes for upstream discussion.

No `plan.md` yet — that waits until upstream review surfaces any structural objections.

## Depends on

**This SDD assumes [#18](https://github.com/Automattic/fosse/pull/18) lands as-is.** That PR establishes the short-form architecture and the `fosse_object_type` projector pattern this spec mirrors for long-form. If #18's architecture changes materially before merge, this spec's pattern assumption needs revisiting. Dependencies are called out in the requirements file's "Pre-Requisite Assumption" section.

## Recommended v1

| # | Option | Status in this spec |
|---|--------|---------------------|
| 1 | Link card (today) | Kept as opt-in alternative |
| 2 | Truncate + link facet | Kept as opt-in alternative |
| 3 | `site.standard.document` record embed | v2 target (when Bluesky renderer ships) |
| 4 | Tweet-storm thread | Rejected |
| 5 | Teaser mini-thread | **Recommended v1 default** |

Option 5 ships a 2-post thread (hook + CTA-with-link) that reads as native to Bluesky. The architecture is structured so a different winner at review is a one-line default-value change — no restructuring needed.

## Open questions (for team discussion, not this PR)

1. 3-post variant default composition — what goes in the middle "takeaway" post if someone opts in via `atmosphere_teaser_thread_posts`.
2. Atomic-write upgrade path — if Bluesky ships a thread-create endpoint or someone invests in client-side CID computation.

## Test plan

- [x] `pnpm run format:check` — clean (docs-only).
- [ ] Team review on the P2 post (link in PR after drafting).
- [ ] Spec revisions based on discussion.
- [ ] `plan.md` added once the spec stabilizes.


kraftbj:
## Pickup notes for tomorrow

After the 2026-04-23 Jim Ray call, this SDD pivoted to Option 5 (teaser mini-thread) as v1 default. Two review passes (plan-executability + AT-Protocol correctness) caught real gaps; confident fixes are in. Flagging below where I made judgment calls so they're visible if anything deserves a rethink before execution starts.

### Decisions locked in

1. **Hook-post cut** = sentence boundary ≤ 280 graphemes (last `.`/`!`/`?`, allowing trailing close-quote/bracket). Word boundary is the fallback when no sentence break exists in the window; mid-word is a last resort for single very long words. Intermediate body-to-body cuts in a future 3+-post variant can be word boundary — sentence-boundary rule applies only to the final body post before the CTA.
2. **CTA copy** = ` __( 'Continue reading: %s', 'atmosphere' ) ` with permalink — translator-aware from day one.
3. **Borderline post** (body ≤ 300 graphemes, titled) = still writes a 2-post thread; no auto-short-circuit to `'truncate-link'`.
4. **Rollback observability** = `error_log` + `WP_Error` return. No admin-UI surface for v1. Partial-meta writes after each successful create are the crash anchor.
5. **Legacy comms** on default flip = `readme.txt` changelog only (FOSSE is pre-1.0). No admin notice, no deprecation window.
6. **Post-meta shape** = single `_atmosphere_bsky_thread_records` array of `{ uri, cid, tid }` triples. Parallel arrays rejected (drift risk; CID-stored-now enables future `swapRecord` without migration).
7. **Crash-recovery** = write partial meta after each successful `applyWrites` create, not once at the end. Eliminates the orphan-records window if PHP fatals mid-thread.
8. **Projector default** = unset / unknown `fosse_long_form_strategy` coerces to `'teaser-thread'`. Opinionated divergence from `Object_Type`'s pass-through.
9. **Linear split** under DOTCOM-16810 = umbrella, no per-task children for v1.
10. **Draft PR first** (Task 1) = open upstream PR as draft as step 1 so Matthias's review window runs in parallel with FOSSE-side tasks. Ready-for-review after Commit 5.
11. **Upstream default for `atmosphere_long_form_composition`** stays `'link-card'`. FOSSE's opinion (thread default) lives entirely in FOSSE's projector. If the Atmosphere plugin team wants to change the upstream default later, that's their call to make separately.
12. **User-set excerpt as hook source.** If `$post->post_excerpt` is non-empty, `build_teaser_thread()` uses it as the hook (user-curated wins over machine body-prefix truncation). Still clamped to 300 graphemes as a safety floor.
13. **Empty / whitespace-only body + no excerpt.** Falls back to `'link-card'` composition for that post only (site option unchanged); notice + log records the degradation so ops can tell it apart from intentional `'link-card'` configuration. Threshold: < 10 graphemes of rendered plain text.
14. **`atmosphere_transform_bsky_post` fires per thread record**, not once per WP post. Consistent treatment across the thread. Behavior change for existing consumers documented in upstream changelog; consumers who need single-invocation semantics can branch on record shape.

### Still genuinely open (not v1 blockers)

- **3-post variant default composition.** v1 ships 2-post; the 3-post path is filter-only via `atmosphere_teaser_thread_posts`. What goes in the middle "takeaway" post is a v1.x decision.
- **Atomic-write upgrade path.** If Bluesky ships a thread-create endpoint or we invest in client-side CID computation, the thread write path gets simpler. Keep the code narrow enough to swap.

Latest commit: `9684cef` — SDD: excerpt-as-hook, empty-body fallback, per-record filter semantics.

RCowles (dismissed):
Architecture is sound, the option analysis is thorough, and the upstream-first split is clean. A few non-blocking observations:

**PR body is stale.** Still lists Option 2 (truncate-link) as the recommended v1, but the spec switched to Option 5 (teaser mini-thread) post-Jim call. Worth updating so the PR summary matches what a reviewer reads in the spec.

**The pivot to Option 5 makes sense.** The Bluesky intro confirmed standard.site rendering is months out with multiple iterations. Truncate-link was a credible short bridge, but as a long-lived default it undersells the product. Teaser thread is a better "this is how FOSSE publishes" story for the end-of-month demo.

**Delete gap in implementation notes.** `on_before_delete` only captures root TID, orphaning thread replies on force-delete. Flagged in the notes but worth fixing before the upstream PR leaves draft — orphaned posts on Bluesky are a bad look.

**`createdAt` behavior change.** Link-card now uses synced-time instead of post date. Subtle breaking change for existing users. Good that it's documented, but should get explicit upstream review rather than shipping quietly.

**2-post default is right for v1.** The 3-post takeaway variant is interesting but adds composition complexity. Let real usage inform whether it's worth adding.

**Projector pattern maps cleanly.** `fosse_long_form_strategy` → `atmosphere_long_form_composition` is the same shape as `fosse_ap_actor_mode` → `activitypub_actor_mode`. Consistency is good.

kraftbj:
Updated the PR body to match the current state of the spec — Option 5 (teaser mini-thread) as v1 default, not Option 2.

Also added your two flags (delete orphaning thread replies on force-delete, and the `createdAt` behavior change for link-card) to the spec under a new "Review Notes" section so they're tracked for follow-up before the upstream PR leaves draft.

PR #25, opened 2026-04-22
SDD: onboarding-setup-ux (requirements + spec + plan)

RCowles:
Add spec-driven development documents for the Onboarding & Setup UX epic (DOTCOM-16793). Covers the unified FOSSE admin menu, inline ActivityPub configuration, Bluesky OAuth bridge via Atmosphere, the connection-status dashboard, and the extensible `Connection_Provider` architecture for future protocol integrations.


sdd/onboarding-setup-ux/implementation-notes.md
```diff
@@ -0,0 +1,52 @@
+# Implementation Notes — Onboarding & Setup UX
```
kraftbj: Quick sanity check on sequencing: this PR only contains the SDD docs — no code, no tests, no `.phpcs.xml.dist` change. But implementation-notes.md reads as if the implementation already landed (19 new tests, deviations from spec, exact line-for-line file descriptions). My guess is you built this locally and pushed the docs first, planning a follow-up PR for the code — can you confirm? Just want to make sure reviewers know whether to read this as "what was built" (evaluate against code) or "what will be built" (evaluate as forward-looking). If it's the latter, might be worth renaming to `decisions.md` / `notes.md` until the code PR lands, or merging this after the code lands so the two stay in sync.

Caveat on my other comments: I've reviewed requirements.md carefully and skimmed plan.md / implementation-notes.md, but I haven't done a deep pass on spec.md yet — will do that once we've aligned on the direction questions below.

sdd/onboarding-setup-ux/spec.md
```diff
@@ -0,0 +1,141 @@
...
+              ├── admin_menu @ 9   → add_menu_page('fosse') + sub-pages
+              ├── admin_menu @ 99  → remove_submenu_page (AP + Atmosphere)
+              └── admin_enqueue_scripts → fosse-admin CSS
+```
+
+Setup_Page and Status_Page iterate `Connection_Provider_Registry::get_providers()` and call `render_setup_section()` / `render_status_card()` on each.
+
+### OAuth Bridge (Bluesky_Provider)
+
+The trickiest part. Atmosphere's `Client::redirect_uri()` returns `admin_url('options-general.php?page=atmosphere')` with no filter. The OAuth callback handler only fires when `$_GET['page'] === 'atmosphere'`.
```
kraftbj: Per the upstream-first policy in AGENTS.md (landed in #23), I'd like to plan this differently. The two gaps we're working around here — `Client::redirect_uri()` being unfilterable and `handle_oauth_callback()` not persisting its success message across the redirect — are both post-type-agnostic: any consumer of Atmosphere running its own admin surface would want these. That puts them squarely in the "fix upstream, consume via sync-bundled" bucket.

Proposal: open PRs against wordpress-atmosphere for (a) a filter on `redirect_uri()` and (b) transient-persisted settings errors on the connect path (matching what disconnect already does). Upstream cadence on AT has been fast — we should be able to land those and re-sync before this FOSSE work ships. That lets `Bluesky_Provider` do the normal thing (set its own redirect_uri, read a standard transient) instead of hooking `wp_redirect` to intercept an Atmosphere URL we don't own.

Worth checking whether anything similar applies to `AP_Provider` (the inline AP settings write). The post-type-agnostic test: does any consumer of the AP plugin outside FOSSE want the same filter/hook? If yes, upstream it.

Happy to take the upstream PRs if you'd rather stay focused on the FOSSE-side work.

sdd/onboarding-setup-ux/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+- **DOTCOM-16800** — Unified FOSSE setup entry point
+- **DOTCOM-16801** — Bluesky handle/DID setup done inside WordPress
+- **DOTCOM-16802** — ActivityPub setup simplification + default-hide advanced settings
+- **DOTCOM-16803** — Unified connection-status dashboard
+
+## Requirements
+
+1. **Single top-level FOSSE admin menu** replacing the need to find AP under Settings and Atmosphere under Settings separately. Two sub-pages: Setup and Status.
+2. **Hide bundled plugins' native admin submenus** from the Settings menu so users are not confused by duplicate surfaces. Pages remain accessible by direct URL for power users.
+3. **Inline ActivityPub configuration** on the Setup page covering the essential settings (actor mode, post types to federate, fediverse address display). Advanced AP settings (relays, blocklists, authorized fetch, hashtags, OpenGraph, etc.) linked out to the original AP settings page.
```
kraftbj: Question on whether we should be picking the "essential vs advanced" line this early. The AP plugin exposes a lot of settings because different site owners genuinely care about different things (blocklists matter to some, relays to others, authorized fetch to the paranoid). If we declare three things essential and everything else "advanced → click out," we're making a product call on behalf of users we haven't talked to yet.

Alternative framing: start the Setup page with the absolute minimum needed to go from zero to federating (maybe just actor mode + post types, since those gate the core behavior), ship it, and let real usage tell us which "advanced" settings get enough clicks to pull into the inline surface. Easier to expand than contract.

Not blocking — just want to flag that this requirement currently reads as a product decision, not a user-research-backed one. If we're confident in the three-item split, call out why in the spec.

sdd/onboarding-setup-ux/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+4. **Bluesky connection via OAuth inside WordPress.** Users enter their Bluesky handle on the FOSSE Setup page, authorize via Atmosphere's OAuth flow (PKCE + DPoP + PAR), and return to the FOSSE Setup page — never needing to visit bsky.app or the Atmosphere settings page. Disconnect also handled from FOSSE.
```
kraftbj: Small reality check: users still have to configure their domain as their Bluesky handle on bsky.app (DNS TXT or `/.well-known` file, then "Change Handle" in the Bluesky app). The OAuth flow doesn't do that for them — it just authorizes an already-configured handle. So "never needing to visit bsky.app" is true only for users who already set up their custom handle.

Probably out of scope for this epic, but worth acknowledging in the doc as a known friction point so we don't claim a smoother UX than we actually deliver. Could be a future FOSSE feature (verify DNS, generate `.well-known` file, guide the user through Bluesky's handle-change UI), or at minimum a help link in the Setup page explaining the prerequisite.

sdd/onboarding-setup-ux/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+5. **Unified connection-status dashboard** showing both protocols on one screen: AP configuration state, Bluesky connection state (handle, DID, PDS), auto-publish toggle, token health, errors, and action buttons (re-auth, disconnect).
+6. **Extensible provider architecture** so future protocol integrations (Leaflet.pub, Threads, etc.) can register as new providers without restructuring the UI. Third-party plugins should also be able to register providers.
+
+## Constraints
+
+- Self-hosted WordPress plugin first. No wpcom dependencies or Calypso APIs.
+- PHP 8.2+, WP 6.9+ (matching FOSSE's existing floor).
+- Bundled plugins (`bundled/activitypub/`, `bundled/atmosphere/`) must NOT be modified. All integration through their public APIs, hooks, and options.
```
kraftbj: Agreed on the no-modification rule for `bundled/`, but want to pair it with the upstream-first policy from AGENTS.md: if we hit a gap in AP or Atmosphere's public API during this work (missing filter, unfilterable URL, non-persistent notice, etc.), the answer is an upstream PR, not a FOSSE-side shim. AP/AT release cadence is fast enough that we can land upstream and re-sync before shipping this epic.

Suggest adding a sentence: "If integration requires a hook or API that doesn't exist upstream, add it via a PR to wordpress-activitypub or wordpress-atmosphere and re-sync, rather than working around the gap in FOSSE." This locks in the policy at the requirements level so reviewers on future FOSSE PRs can point to it.

sdd/onboarding-setup-ux/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+- `bundled/atmosphere/includes/wp-admin/class-admin.php:325-363` — OAuth callback handler (`handle_oauth_callback`), checks `$_GET['page'] === 'atmosphere'`, calls `wp_safe_redirect` to `options-general.php?page=atmosphere&connected=1` on success.
+- `bundled/atmosphere/includes/oauth/class-client.php:41-42` — `redirect_uri()` returns `admin_url('options-general.php?page=atmosphere')`, hardcoded with no filter.
+- `bundled/atmosphere/includes/oauth/class-client.php:54` — `authorize($handle)` is `public static`, returns `string|\WP_Error`.
+- `bundled/atmosphere/includes/oauth/class-client.php:500` — `disconnect()` is `public static`, deletes `atmosphere_connection` option.
+- `bundled/atmosphere/includes/functions.php:109` — `is_connected()` checks `!empty($conn['access_token']) && !empty($conn['did'])`.
+
+## Open Questions
+
+- **AP settings ownership.** We write directly to `activitypub_actor_mode` and `activitypub_support_post_types` via `update_option()`. If AP's settings registration changes upstream, our saves could conflict. Acceptable for MVP since the bundled version is pinned, but worth monitoring.
```
kraftbj: Open question on whether this should work differently: instead of overwriting AP's options directly, could FOSSE store its own options (e.g. `fosse_ap_actor_mode`) and register `pre_option_activitypub_actor_mode` / `option_activitypub_actor_mode` filters that project FOSSE's value down to AP at read time? That gives us:

1. No write-time coupling to AP's option schema (if AP adds sanitize callbacks or changes the option name, we don't silently bypass them).
2. Clear ownership story — FOSSE's Setup page owns `fosse_*`, AP's advanced page owns `activitypub_*`, and the filter layer keeps them in sync when FOSSE is in control.
3. Easy to unset — delete the FOSSE option, AP's own stored value takes over again. Makes "stop letting FOSSE manage this" a one-line op.

The downside is a tiny runtime cost and one more layer of indirection. For two options it's probably fine; worth deciding the pattern now before we add more.

Thoughts? If this shape makes sense, it would also apply to any future Atmosphere settings we surface.

kraftbj (commented):
Partial review focused on `requirements.md` and `implementation-notes.md`; haven't done a deep pass on `spec.md` yet. Leaving inline comments below for discussion before I dive further.

sdd/onboarding-setup-ux/implementation-notes.md
RCowles: Good catch. This is forward-looking. I built it locally first to validate the approach, then pushed the SDD docs ahead of the code so we could align on the shape before the implementation PR lands. I'll reword implementation-notes.md to make that clear, framing the deviations and notes as planned decisions rather than post-facto documentation.

sdd/onboarding-setup-ux/spec.md
RCowles: This is the right call. I was working around the gap because I assumed we'd need to move fast, but if upstream cadence is fast enough to land before we ship, fixing it at the source is strictly better. Let's open PRs against wordpress-atmosphere for (a) a redirect_uri filter and (b) transient-persisted settings errors on connect. Happy to take those, or if you want to since you're closer to the Atmosphere codebase, I'll stay focused on the FOSSE side. Either way, I'll update the spec to reflect the upstream-first approach.

sdd/onboarding-setup-ux/requirements.md
RCowles: Agreed. Actor mode + post types is the right starting line. Those are the two things that gate whether federation actually works. Everything else can wait for real usage data. I'll tighten the requirement to reflect that and remove the "essential vs. advanced" framing.

sdd/onboarding-setup-ux/requirements.md
RCowles: Good reality check. I'll add a note acknowledging that users still need to configure their domain as their Bluesky handle separately (DNS TXT or well-known + Bluesky's "Change Handle" flow). We can flag it as a future FOSSE feature and add a help link in the Setup page for now.

sdd/onboarding-setup-ux/requirements.md
RCowles: Added. "If integration requires a hook or API that doesn't exist upstream, add it via a PR to wordpress-activitypub or wordpress-atmosphere and re-sync, rather than working around the gap in FOSSE."

sdd/onboarding-setup-ux/requirements.md
RCowles: I like this pattern a lot. Cleaner ownership, no bypass risk, and easy to unset. I'll update the spec to use fosse_ap_* options with pre_option_* filters projecting down to AP. Establishes the right pattern before we add more settings.

sdd/onboarding-setup-ux/spec.md
```diff
@@ -0,0 +1,147 @@
...
+Form POSTs to `admin_post.php?action=fosse_save_ap_settings`. The handler validates against the allowlist of actor mode values and the list of public post types, then calls `update_option()` on the `fosse_ap_*` keys.
+
+### Key Components
+
+| Component | File | Responsibility |
+|---|---|---|
+| `Connection_Provider` | `src/Admin/Connection_Provider.php` | Interface: slug, name, availability, status, setup render, status render, hook registration. |
+| `Connection_Provider_Registry` | `src/Admin/Connection_Provider_Registry.php` | Static registry: register, get_providers, get_provider, reset. |
+| `AP_Provider` | `src/Admin/AP_Provider.php` | ActivityPub setup section (inline config) + status card. Self-registers on `fosse_register_providers`. |
+| `Bluesky_Provider` | `src/Admin/Bluesky_Provider.php` | Bluesky setup section (OAuth connect/disconnect) + status card. Self-registers on `fosse_register_providers`. Depends on upstream Atmosphere filter for redirect_uri. |
```
kraftbj: Small residual wording nit: "OAuth bridge" is a remnant from the old wp_redirect-intercept approach. Now that we're going upstream-first, this row could read something like "Bluesky setup section (OAuth connect/disconnect) + status card. Self-registers on `fosse_register_providers`. Depends on upstream Atmosphere `redirect_uri` filter + persisted-notice transient." Optional.

sdd/onboarding-setup-ux/spec.md
```diff
@@ -0,0 +1,147 @@
...
+| `src/Admin/Bluesky_Provider.php` | new | Bluesky provider implementation with OAuth bridge. |
+| `src/Admin/Menu.php` | new | Admin menu registration and bundled-menu suppression. |
+| `src/Admin/Setup_Page.php` | new | Setup page controller. |
+| `src/Admin/Status_Page.php` | new | Status dashboard controller. |
+| `src/Admin/templates/setup-page.php` | new | Setup page HTML template. |
+| `src/Admin/templates/status-page.php` | new | Status dashboard HTML template. |
+| `src/Admin/assets/css/admin.css` | new | Admin styles. |
+| `tests/php/Admin/Connection_Provider_RegistryTest.php` | new | Registry unit tests. |
+| `tests/php/Admin/AP_ProviderTest.php` | new | AP provider unit tests. |
+| `tests/php/Admin/Bluesky_ProviderTest.php` | new | Bluesky provider unit tests (redirect intercept, status data). |
```
kraftbj: Same stale-wording nit: "redirect intercept, status data" describes tests for the old approach. After the Task 4 rewrite this row should read something like "Bluesky provider unit tests (redirect URI filter integration, persisted-notice read, status data)." Optional.

kraftbj (commented):
Second pass after the force-push. `requirements.md` and `spec.md` both look good to go — every review point was addressed cleanly. `plan.md` still needs a full refresh to match the updated spec (stale wp_redirect intercept, missing option-projection steps, no upstream-PR task, outdated "Backfill" framing for Task 9). Copilot flagged the wp_redirect piece too — fully agree with that comment. Also agree with Copilot's suggestion to rename the `implementation-notes.md` H1 / intro so it doesn't mismatch the filename. Two tiny residual nits on `spec.md` inline below — optional, but while you're in there.

Resolving the threads that have been substantively addressed by the rewrite.

sdd/onboarding-setup-ux/implementation-notes.md
```diff
@@ -0,0 +1,42 @@
+# Planned Decisions — Onboarding & Setup UX
+
+These are forward-looking notes documenting decisions made during the design phase. They will be updated as implementation-notes once the code PR lands.
```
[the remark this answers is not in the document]
kraftbj: +1 to this. Either rename the file (e.g. `design-notes.md` or `planned-decisions.md`) or keep the filename and adopt your suggested H1/intro so readers immediately understand this is pre-implementation. Leaving this thread open until Ryan decides which direction.

sdd/onboarding-setup-ux/spec.md
RCowles: @copilot rephrase the wording to address Kraft's wording nit (he's right!)

sdd/onboarding-setup-ux/spec.md
RCowles: @copilot rephrase this one per the nit as well

sdd/onboarding-setup-ux/implementation-notes.md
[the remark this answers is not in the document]
RCowles: @copilot rename this file to `planned-decisions.md`.

sdd/onboarding-setup-ux/requirements.md
```diff
@@ -0,0 +1,69 @@
...
+- **DOTCOM-16793** — 1. Onboarding & Setup UX (parent epic, assigned to Ryan)
+- **DOTCOM-16800** — Unified FOSSE setup entry point
+- **DOTCOM-16801** — Bluesky handle/DID setup done inside WordPress
+- **DOTCOM-16802** — ActivityPub setup simplification + default-hide advanced settings
+- **DOTCOM-16803** — Unified connection-status dashboard
+
+## Requirements
+
+1. **Single top-level FOSSE admin menu** replacing the need to find AP under Settings and Atmosphere under Settings separately. Two sub-pages: Setup and Status.
+2. **Hide bundled plugins' native admin submenus** from the Settings menu so users are not confused by duplicate surfaces. Pages remain accessible by direct URL for power users.
```
kraftbj: Expand the scope here — not just the Settings submenu. Verified against `bundled/activitypub/includes/wp-admin/class-menu.php`:

- Line 28 — AP's Settings submenu (`activitypub`, under `options-general.php`). Already handled.
- Lines 42-49 — AP registers a **top-level Dashboard page** (`activitypub-social-web`) when `activitypub_reader_ui` is on. Not handled.
- Lines 57-79 — AP registers two **Users submenus** (`activitypub-followers-list`, `activitypub-following-list`). Not handled.

FOSSE is claiming the unified admin surface for federation setup, so all of these should be hidden for consistency. We can add FOSSE-owned equivalents (follower lists, dashboard) later when we have a story for them. Better to start with one clean surface and build up than ship with three disjoint entry points.

Suggested requirement rewording: **"Hide all bundled-plugin admin entries (Settings submenu, top-level Dashboard page, Users submenus) so FOSSE is the single admin surface for federation setup. Pages remain accessible by direct URL for power users."**

sdd/onboarding-setup-ux/plan.md
```diff
@@ -0,0 +1,119 @@
...
+  - `composer dump-autoload` succeeds.
+  - `composer run-script lint-php` passes on new files.
+- **Depends on**: none
+
+### Task 2: Create Menu.php + wire up in fosse.php
+- **Files**: `src/Admin/Menu.php`, `fosse.php`
+- **Do**:
+  1. Create `Menu` class with `register()` static method that fires `fosse_register_providers`, calls `register_hooks()` on available providers, hooks `admin_menu` at priority 9 (register menu) and 99 (hide bundled menus), and hooks `admin_enqueue_scripts` for CSS.
+  2. `add_menu_page('fosse')` with `dashicons-share` icon. Two submenu pages: Setup (slug `fosse`) and Status (slug `fosse-status`).
+  3. `remove_submenu_page('options-general.php', 'activitypub')` and `remove_submenu_page('options-general.php', 'atmosphere')` at priority 99.
```
kraftbj: Task 2 step 3 should expand with the broader menu scope. The current plan only removes the two Settings submenu entries:
```
remove_submenu_page('options-general.php', 'activitypub')
remove_submenu_page('options-general.php', 'atmosphere')
```

Add removals for AP's top-level Dashboard and Users submenus:
```
remove_menu_page('activitypub-social-web')
remove_submenu_page('users.php', 'activitypub-followers-list')
remove_submenu_page('users.php', 'activitypub-following-list')
```

(Pages remain registered, so direct-URL access still works for power users.)

Atmosphere's admin is a single Settings submenu, so no additional removal there.

Verify step should add: "Playground: no AP or Atmosphere entries appear in Settings, Users, or the top-level menu."

sdd/onboarding-setup-ux/planned-decisions.md
```diff
@@ -0,0 +1,42 @@
...
+### Upstream-first OAuth integration (not wp_redirect intercept)
+- **Initial approach**: Hook `wp_redirect` to intercept Atmosphere's post-callback redirect and rewrite it to the FOSSE page.
+- **Revised approach**: Open upstream PRs against wordpress-atmosphere for (a) a `redirect_uri` filter and (b) transient-persisted settings errors on connect. Consume via sync-bundled once landed.
+- **Reason**: Per the upstream-first policy in AGENTS.md. Both gaps are post-type-agnostic (any consumer of Atmosphere running its own admin surface would want these), so they belong upstream, not as FOSSE-side shims.
+
+### Provider `register_hooks()` called from `Menu::register()`
+- **Decision**: `Menu::register()` fires `fosse_register_providers`, then iterates registered providers and calls `register_hooks()` on each. Provider hooks (like `admin_post_*` handlers) are only registered in admin context.
+- **Reason**: Avoids registering admin-post handlers on front-end requests. The `is_admin()` guard in `fosse.php` already limits this to admin context, but having `Menu::register()` control the lifecycle is cleaner.
+
+## Known Limitations (Expected)
```
kraftbj, line 27: Add a deferral note under "Known Limitations (Expected)" covering the deactivation/deletion story:

> **Deactivation/deletion behavior is deferred.** This epic does not define what happens to `fosse_ap_*` options or menu state when FOSSE is deactivated or deleted, nor how FOSSE behaves if ActivityPub or Atmosphere are also installed as standalone plugins. Tracked separately in [DOTCOM-16865](https://linear.app/a8c/issue/DOTCOM-16865/deactivation-and-deletion-handling-for-fosse-and-bundled-plugins) so the decision happens before production distribution.

The option-projection pattern (FOSSE stores `fosse_ap_*`, projects via `pre_option_activitypub_*`) is clean for the "FOSSE is active" case but needs a companion story for lifecycle events. Not blocking for MVP.

kraftbj (commented):
Two follow-ups from a comprehensive review pass (full inventory below; the rest all checked out — claims about `bundled/` source verified accurate, upstream-first honored, structural parity with `sdd/bundled-backends/`, AGENTS.md compliance clean).

1. **AP menu suppression scope** — plan currently hides only AP's Settings submenu. AP also registers a top-level Dashboard page (`activitypub-social-web`) and two Users submenus (`activitypub-followers-list`, `activitypub-following-list`). Since FOSSE is claiming the full admin surface for federation setup, those should all be hidden too — we can build FOSSE-owned equivalents later. Start clean. Details inline.
2. **Deactivation/deletion story** — deliberately deferred out of this epic. Tracked in DOTCOM-16865. A one-line note under "Known Limitations" in `planned-decisions.md` pointing at that issue is enough.

Everything else is fine. Ship once these are in.

PR #27, opened 2026-04-23
Add FOSSE admin UI with Connection_Provider architecture

RCowles:
## Summary

Implements the FOSSE admin UI per the [onboarding-setup-ux SDD](https://github.com/Automattic/fosse/blob/trunk/sdd/onboarding-setup-ux/spec.md) (merged in #25). Replaces the bundled plugins' fragmented settings pages with a single FOSSE-owned admin surface.

- **Connection_Provider interface + registry** — extensible abstraction so future protocols (Bluesky, Leaflet.pub, etc.) register as providers without restructuring the UI. Third-party plugins can also register providers via the `fosse_register_providers` hook.
- **Top-level FOSSE menu** with Setup and Status sub-pages. Hides all bundled-plugin admin entries (AP Settings, Dashboard, Users submenus; Atmosphere Settings) at priority 99. Pages remain accessible by direct URL for power users.
- **AP_Provider with option projection** — inline configuration for actor mode and post types. Stores `fosse_ap_*` options and projects them to AP via `pre_option_*` filters (priority 20, so AP's constant-based overrides at priority 10 take precedence). Status card shows actor mode, post types, fediverse address, and follower count.
- **PHPCS `src/` exclusion** from WordPress.Files.FileName sniff for PascalCase namespaced files.

### What's not in this PR

- **Bluesky_Provider (Task 5)** — blocked on upstream Atmosphere PRs for a `redirect_uri` filter and transient-persisted connect notices. See `sdd/onboarding-setup-ux/planned-decisions.md` for details.
- **Bluesky_Provider tests** — will ship with Task 5.
- **Onboarding wizard** — shipping in a separate stacked PR.
- **SDD docs update (Task 10)** — will follow once implementation is complete.

### SDD plan task mapping

| Task | Status | Commit |
|------|--------|--------|
| 1: Connection_Provider interface + registry | Done | `3efb10d` |
| 2: Menu + fosse.php wiring | Done | `c7eb049` |
| 3: AP_Provider + option projection | Done | `72c86e1` |
| 4: Upstream Atmosphere PRs | Blocked (external) | — |
| 5: Bluesky_Provider | Blocked on Task 4 | — |
| 5.5: Onboarding wizard | Separate PR | — |
| 6: Status_Page | Done | `c7eb049` |
| 7: Admin CSS | Done | `c7eb049` |
| 8: Tests (Registry + AP_Provider) | Done | `3efb10d`, `72c86e1` |
| 9: PHPCS exclusion | Done | `08c773f` |
| 10: SDD docs update | After implementation | — |

### Deviation from SDD

Menu suppression now also removes `activitypub-blocked-actors-list` and the Extra Fields CPT redirect under Users, which were not listed in the original SDD plan (@kraftbj's review only found the entries in `class-menu.php` lines 57-79). Discovered during Playground testing.

## Test plan

- [ ] `composer run-script test-php` — 39 tests pass (Registry: 5, AP_Provider: 21, existing: 13)
- [ ] `composer run-script lint-php` — clean
- [ ] `pnpm run format:check` — clean
- [ ] Playground: FOSSE menu appears with Setup and Status sub-pages
- [ ] Playground: no AP or Atmosphere entries in Settings, Users, or top-level menu
- [ ] Playground: AP Setup section renders with actor mode, post types, fediverse address
- [ ] Playground: saving AP settings persists and reflects on AP's own settings page (option projection)
- [ ] Playground: Status page shows AP card with green indicator and correct data

Refs: DOTCOM-16793, DOTCOM-16800, DOTCOM-16802, DOTCOM-16803


src/Admin/class-menu.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @return void
+	 */
+	public static function enqueue_styles( string $hook_suffix ): void {
+		if ( ! str_starts_with( $hook_suffix, 'toplevel_page_fosse' ) && ! str_starts_with( $hook_suffix, 'fosse_page_' ) ) {
+			return;
+		}
+
+		wp_enqueue_style(
+			'fosse-admin',
+			plugins_url( 'src/Admin/assets/css/admin.css', dirname( __DIR__, 2 ) . '/fosse.php' ),
```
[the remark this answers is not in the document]
RCowles: Not a bug — `dirname(__DIR__, 2)` from `src/Admin/Menu.php` resolves to the repo root (up two levels from `src/Admin`), so the second arg is `<root>/fosse.php`, which is correct. The suggested `dirname(__DIR__, 3)` would go one level too high.

src/Admin/templates/setup-page.php
```diff
@@ -0,0 +1,32 @@
...
+			$provider->render_setup_section();
+		}
+	}
+	?>
+
+	<?php if ( empty( $providers ) ) : ?>
+		<div class="notice notice-warning">
+			<p><?php esc_html_e( 'No federation providers are available. Ensure ActivityPub and Atmosphere are installed.', 'fosse' ); ?></p>
+		</div>
+	<?php endif; ?>
```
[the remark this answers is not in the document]
RCowles: Technically true, but providers only register when `is_available()` returns true (see `AP_Provider::register_provider()`), so a provider in the registry is always available. The `is_available()` check in the template loop is defensive, not load-bearing. The status-page template is more defensive about this — not worth aligning for MVP but noted.

src/Admin/AP_Provider.php
```diff
@@ -0,0 +1,356 @@
...
+	 *
+	 * @return string Empty string if AP models are unavailable.
+	 */
+	private function get_fediverse_address(): string {
+		if ( class_exists( '\Activitypub\Model\Blog' ) ) {
+			$blog = new \Activitypub\Model\Blog();
+			return $blog->get_webfinger();
+		}
+
+		return '';
```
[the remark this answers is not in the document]
RCowles: Good catch — fixed in 82e65a8. Address now uses the user webfinger in actor mode, blog in blog mode, and user-with-blog-fallback in actor_blog mode.

src/Admin/AP_Provider.php
```diff
@@ -0,0 +1,356 @@
...
+	private function render_follower_count_row(): void {
+		if ( ! class_exists( '\Activitypub\Collection\Followers' ) ) {
+			return;
+		}
+
+		$count = \Activitypub\Collection\Followers::count( get_current_user_id() );
+		?>
+		<tr>
+			<td><?php esc_html_e( 'Followers', 'fosse' ); ?></td>
+			<td><?php echo esc_html( number_format_i18n( $count ) ); ?></td>
```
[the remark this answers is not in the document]
RCowles: Good catch — fixed in 82e65a8. Follower count now uses `Actors::BLOG_USER_ID` in blog mode and shows both author and blog counts in actor_blog mode.

src/Admin/AP_Provider.php
```diff
@@ -0,0 +1,405 @@
...
+		check_admin_referer( 'fosse_save_ap_settings' );
+
+		// Sanitize actor mode against allowlist.
+		$mode = sanitize_text_field( wp_unslash( $_POST['fosse_ap_actor_mode'] ?? '' ) );
+		if ( in_array( $mode, self::ACTOR_MODES, true ) ) {
+			update_option( 'fosse_ap_actor_mode', $mode );
+		}
+
+		// Sanitize post types against registered public types.
+		$submitted   = array_map( 'sanitize_text_field', wp_unslash( $_POST['fosse_ap_support_post_types'] ?? array() ) );
```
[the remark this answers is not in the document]
RCowles: Good catch — fixed in 34bfd48. Used `(array)` cast to safely handle malformed input.

src/Admin/AP_Provider.php
```diff
@@ -0,0 +1,405 @@
...
+		$mode = get_option( 'activitypub_actor_mode', 'actor' );
+
+		if ( 'blog' !== $mode && class_exists( '\Activitypub\Model\User' ) ) {
+			$user = \Activitypub\Model\User::from_wp_user( get_current_user_id() );
+			if ( $user && ! is_wp_error( $user ) ) {
+				return $user->get_webfinger();
+			}
+		}
+
+		if ( class_exists( '\Activitypub\Model\Blog' ) ) {
```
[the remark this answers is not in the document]
RCowles: Valid — fixed in 2314839. Blog fallback now only applies in actor_blog mode. Pure actor mode returns empty string if the user actor is unavailable.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,405 @@
...
+	 * @return string|false
+	 */
+	public function project_actor_mode( $pre ) {
+		if ( false !== $pre ) {
+			return $pre;
+		}
+
+		$fosse_value = get_option( 'fosse_ap_actor_mode' );
+		if ( false !== $fosse_value ) {
+			return $fosse_value;
```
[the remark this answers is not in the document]
RCowles: FOSSE is the only writer of `fosse_ap_actor_mode` and `handle_save()` validates against the allowlist before storing. Re-validating on every `get_option` read adds overhead for a scenario that requires direct DB manipulation to trigger. Not worth the runtime cost.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,405 @@
...
+	 * @param array|false $pre The pre-get option value.
+	 * @return array|false
+	 */
+	public function project_post_types( $pre ) {
+		if ( false !== $pre ) {
+			return $pre;
+		}
+
+		$fosse_value = get_option( 'fosse_ap_support_post_types' );
+		if ( false !== $fosse_value ) {
```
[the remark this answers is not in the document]
RCowles: Same reasoning as the actor mode comment. handle_save() validates against get_post_types before storing. Re-querying the post type registry on every get_option read is unnecessary overhead for internal code.

src/Admin/templates/status-page.php
```diff
@@ -0,0 +1,51 @@
...
+
+// phpcs:disable VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable -- $providers set by Status_Page::render().
+$available = array_filter(
+	$providers,
+	static fn( $p ) => $p->is_available()
+);
+$connected = array_filter(
+	$available,
+	static fn( $p ) => ! empty( $p->get_status()['connected'] )
+);
```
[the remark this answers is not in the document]
RCowles: Premature optimization. Both providers get_status() methods are trivial get_option() calls. Not worth adding a caching layer for two cheap reads.

RCowles:
@kraftbj Ready for review. This is the architectural foundation for the onboarding UX, not the polished UI itself. The goal here is getting the right pieces in place: provider abstraction, option projection, menu suppression, and page shells that everything else builds on top of.

The Copilot review caught a few real issues (actor-mode-aware fediverse address and follower count, input type guard) which are fixed. It also suggested some overly defensive validation on the projection filters that I declined with reasoning inline.

Bluesky_Provider (Task 5) is blocked on upstream Atmosphere PRs. I can take a first pass at this.

fosse.php
```diff
@@ -120,3 +120,17 @@ static function () {
...
+
+/*
+ * Admin UI: FOSSE setup and status pages.
+ *
+ * Each provider self-registers on the 'fosse_register_providers' action,
+ * and Menu::register() fires that action, adds the admin menu, and
+ * enqueues styles. Bundled-plugin admin entries are hidden so FOSSE is
+ * the single admin surface for federation setup.
+ */
+if ( is_admin() && class_exists( \Automattic\Fosse\Admin\Menu::class ) ) {
```
kraftbj: **Critical — option projection is silently skipped outside wp-admin.**

This `is_admin()` guard wraps both `AP_Provider::init()` and `Menu::register()`. `Menu::register()` is where `do_action('fosse_register_providers')` fires and where each provider's `register_hooks()` gets called — and `register_hooks()` is where the `pre_option_activitypub_actor_mode` and `pre_option_activitypub_support_post_types` filters get added (`src/Admin/AP_Provider.php:228-229`).

So on any request where `is_admin()` is false — WebFinger, `/wp-json/activitypub/*`, outbox, actor profiles, `wp-cron.php`, AP's `init`-time post-type registration, the federation transformer — the projection filters never attach. AP falls through to its own stored `activitypub_*` options.

Concrete failure: a user switches actor mode to `blog` in the FOSSE UI, the Status card reflects `blog`, but AP keeps serving per-author actors and federating against its own untouched `activitypub_support_post_types` default. The whole "FOSSE owns settings, AP reads through transparently" promise is broken everywhere except wp-admin.

The fix is to split the bootstrap: provider discovery + `register_hooks()` run unconditionally (probably on `plugins_loaded`), and only the menu/enqueue/page rendering stay behind `is_admin()`. A regression test that asserts the filters attach in a non-admin context would lock this in.

.phpcs.xml.dist
```diff
@@ -37,6 +37,7 @@
 	</rule>
 
 	<rule ref="WordPress.Files.FileName">
+		<exclude-pattern>src/</exclude-pattern>
```
kraftbj: **Filename convention will block WP.com/Simple inclusion.**

Excluding `src/` from `WordPress.Files.FileName` lets the PascalCase files pass lint locally, but Simple enforces WPCS filename rules (lowercase-hyphenated `class-foo.php`) and won't take this as-is. Worth noting the bundled AP plugin under `bundled/activitypub/includes/` already uses `class-activitypub.php`, `class-followers.php`, etc. — the inconsistency is inside FOSSE itself.

PSR-4 is off the table for the same reason. The natural path is [`automattic/jetpack-autoloader`](https://packagist.org/packages/automattic/jetpack-autoloader) which is designed for exactly this constraint and already proven on .com. A hand-rolled `spl_autoload_register` mapping `Automattic\Fosse\Admin\AP_Provider` → `src/Admin/class-ap-provider.php` is the lighter alternative if you'd rather not take the dep.

Doesn't have to land in this PR — but do the rename + drop this exclusion before FOSSE is a serious WP.com candidate.

src/Admin/Menu.php
```diff
@@ -0,0 +1,120 @@
...
+	 *
+	 * @return void
+	 */
+	public static function register(): void {
+		/**
+		 * Fires so Connection_Provider implementations can register themselves.
+		 *
+		 * Providers call Connection_Provider_Registry::register( $this ) here.
+		 */
+		do_action( 'fosse_register_providers' );
```
kraftbj: Provider registration race: the `do_action('fosse_register_providers')` fires as soon as `Menu::register()` runs, and the registry is snapshotted immediately after in the foreach. Any provider whose bootstrap only becomes available on `init` / `admin_init` (e.g. classes not yet autoloaded, or a third-party provider that needs `plugins_loaded` itself) will miss the window and silently never register its `pre_option_*` or `admin_post_*` hooks.

Less urgent than the `is_admin()` issue in `fosse.php:132`, but worth solving at the same time — if you move registration to `plugins_loaded`, make sure third-party providers can still land at `init` if they need to.

src/Admin/class-menu.php
```diff
@@ -0,0 +1,120 @@
...
+		remove_submenu_page( 'options-general.php', 'atmosphere' );
+
+		// AP top-level Dashboard page (gated by activitypub_reader_ui).
+		remove_menu_page( 'activitypub-social-web' );
+
+		// AP Users submenus.
+		remove_submenu_page( 'users.php', 'activitypub-followers-list' );
+		remove_submenu_page( 'users.php', 'activitypub-following-list' );
+		remove_submenu_page( 'users.php', 'activitypub-blocked-actors-list' );
+		remove_submenu_page( 'users.php', esc_url( admin_url( '/edit.php?post_type=ap_extrafield' ) ) );
```
kraftbj, line 91: Suppressing the Extra Fields CPT submenu by matching the rendered `edit.php?post_type=ap_extrafield` URL is fragile — if AP ever changes how it registers that submenu (`show_in_menu`, a different query string order, etc.), this stops matching and the row leaks back with no warning.

Not a blocker since FOSSE ships AP bundled, so you control the coupling. A short `// Mirrors bundled AP registration at …/class-menu.php:98` comment would help the next person spot it during an AP version bump.

src/Admin/AP_Provider.php
```diff
@@ -0,0 +1,417 @@
...
+	 * @return void
+	 */
+	private function render_follower_count_row(): void {
+		if ( ! class_exists( '\Activitypub\Collection\Followers' ) ) {
+			return;
+		}
+
+		$mode    = get_option( 'activitypub_actor_mode', 'actor' );
+		$blog_id = defined( '\Activitypub\Collection\Actors::BLOG_USER_ID' )
+			? \Activitypub\Collection\Actors::BLOG_USER_ID
```
kraftbj: The `defined()` guard falls back to `$blog_id = 0` when the constant is absent. If a future AP version renames or removes `\Activitypub\Collection\Actors::BLOG_USER_ID` but keeps `Followers::count()` callable, you'll end up counting followers for user ID `0` and rendering a bogus number instead of suppressing the row.

Safer pattern: if the constant is missing, skip the blog-count branch entirely (return early from the blog/actor_blog paths), rather than querying with `0`.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,417 @@
...
+						esc_html__( 'Author: %1$s, Blog: %2$s', 'fosse' ),
+						esc_html( number_format_i18n( $user_count ) ),
+						esc_html( number_format_i18n( $blog_count ) )
+					);
+					?>
+				</td>
+			</tr>
+			<?php
+		} else {
+			$count = \Activitypub\Collection\Followers::count( get_current_user_id() );
```
kraftbj, line 366: In pure `actor` mode this renders `Followers::count( get_current_user_id() )` under the site-wide "Followers" label — so on a multi-author site two admins looking at the Status page see different numbers for the same "site." Not a correctness bug, but it misrepresents what's being shown.

Two reasonable paths: relabel to "Your followers" when `$mode === 'actor'`, or sum across users with an AP-enabled role. The relabel is cheaper and honest.

kraftbj (changes requested):
Heads up: this is a **static review** — I haven't user-tested any of this yet, so treat the severity calls as code-reading confidence, not behavior-confirmed. Requesting changes on one real correctness issue; the rest is commentary.

Architecturally the PR lands where it should. The `Connection_Provider` abstraction, option-projection pattern, and bundled-menu suppression all read well; nonces/caps/escaping are clean throughout `handle_save` and the templates; Copilot's four real bugs all got the right fixes (82e65a8 / 34bfd48 / 2314839) and the three declines are defensible. The one must-fix is that projection silently doesn't run outside wp-admin, which defeats the pattern.

## Must fix before merge

1. **`is_admin()` gates the projection filters** — AP reads its own options on WebFinger, outbox, REST, and cron. Inline on `fosse.php:132`. This is the big one.

## Major — needs a plan, not necessarily this PR

2. **PascalCase filenames block WP.com/Simple inclusion.** PSR-4 isn't an option over there; WPCS filenames (`class-foo.php`) are required. Inline on `.phpcs.xml.dist`.

## Minor / follow-up

3. Provider registration race when providers arrive after the `do_action` fires (`Menu.php:30`). Solvable alongside #1.
4. `Extra Fields` submenu suppression matches a rendered URL string; brittle but low-risk (`Menu.php:99`).
5. `BLOG_USER_ID` constant-absence fallback to `0` can surface a bogus row (`AP_Provider.php:377`).
6. In pure `actor` mode the follower-count row shows the viewing admin's count under a site-wide label (`AP_Provider.php:408`).

## Test gaps worth filling

- Projection exercised in a non-admin request context (would have caught #1).
- `handle_save` capability + nonce denial paths, plus allowlist-rejection for actor mode and post types.
- E2E assertion that direct URLs to the suppressed bundled pages still work — the PR description promises this, but nothing pins it.

In short: fix # 1, pick a path on filenames/autoloader (Jetpack autoloader is the obvious one), and the rest can roll into follow-ups.

src/Admin/class-menu.php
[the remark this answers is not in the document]
kraftbj: Agree with Ryan here — Copilot got this one wrong. From `src/Admin/Menu.php`, `__DIR__` is `src/Admin`, so `dirname(__DIR__, 2)` is the repo root and `dirname(__DIR__, 2) . '/fosse.php'` resolves correctly. `dirname(__DIR__, 3)` would climb above the plugin directory. Resolving.

RCowles:
> Bluesky_Provider (Task 5) is blocked on upstream Atmosphere PRs. I can take a first pass at this.

Ref: https://github.com/Automattic/wordpress-atmosphere/pull/33

.phpcs.xml.dist
RCowles: Agreed this needs to happen before WP.com. Two questions to help me pick the right path:

1. Is jetpack-autoloader the strong recommendation here, or would a hand-rolled spl_autoload_register mapping be acceptable? Jetpack autoloader is proven but adds a dep; hand-rolled is lighter but one more thing to maintain.

2. Should we do the rename in this PR (keeps the diff self-contained) or as an immediate follow-up? The rename itself is mechanical — every file in src/ gets class-*.php naming and we drop the PHPCS exclusion. Happy to do it either way.

src/Admin/Menu.php
RCowles: Largely addressed by the Provider_Loader refactor (splitting provider hooks out of Menu). Both first-party providers init in the same file, so there is no race today.

For third-party providers the timing question is real — if boot() fires at plugin include time and a third-party hooks fosse_register_providers on plugins_loaded, they miss it. Worth solving when third-party registration becomes a real use case. Deferring boot() to a plugins_loaded hook at a defined priority would be the fix, but it changes init ordering for everything else in fosse.php and is not worth the risk for two first-party providers.

.phpcs.xml.dist
kraftbj: I think `jetpack-autoloader` is good, just because we do maintain it and it's main to work with multiple WP plugins using the same packages, so it would be trivial if we do something like converting AP/AT stuff to a package that both products consume instead of bundling the plugin directly. I think the weight of the deps is worth it vs having something bespoke IMO.

Might as well do it in this PR, I think.

RCowles: Done in a72228a. Added jetpack-autoloader, renamed all src/ files to class-*.php / interface-*.php, dropped the PHPCS exclusion, updated fosse.php to load autoload_packages.php. Build-zip sanity check updated too.

fosse.php
RCowles: Fixed in 65dc997. Provider bootstrap is now in Provider_Loader::boot() which runs unconditionally in fosse.php — no is_admin() gate. Menu::register() stays admin-only for menu pages, CSS, and bundled-menu suppression. Regression test in Provider_LoaderTest verifies projection works without is_admin() context.

src/Admin/AP_Provider.php
RCowles: Fixed in d178598. Blog-count branch now skips entirely when the constant is absent instead of falling back to 0.

src/Admin/class-menu.php
RCowles: Added source reference comment in d178598.

kraftbj (dismissed):
Architecture is solid. The provider abstraction, option-projection pattern, and the Provider_Loader split (fixing the is_admin gate) are all well-done. Copilot's real bugs got fixed correctly, the jetpack-autoloader switch was the right call, and test coverage on the projection/registry/loader paths is good.

Requesting changes on one correctness issue (actor mode change doesn't propagate to federation) and a couple of UX honesty items. The rest is follow-up or suggestions. Inline comments have severity labels.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,418 @@
...
+		if ( ! current_user_can( 'manage_options' ) ) {
+			wp_die( esc_html__( 'You do not have permission to do this.', 'fosse' ) );
+		}
+
+		check_admin_referer( 'fosse_save_ap_settings' );
+
+		// Sanitize actor mode against allowlist.
+		$mode = sanitize_text_field( wp_unslash( $_POST['fosse_ap_actor_mode'] ?? '' ) );
+		if ( in_array( $mode, self::ACTOR_MODES, true ) ) {
+			update_option( 'fosse_ap_actor_mode', $mode );
```
kraftbj: **BEFORE MERGE:** Writing `fosse_ap_actor_mode` and projecting it at read time works for option reads, but AP schedules blog profile updates off `add_option_activitypub_actor_mode` / `update_option_activitypub_actor_mode` hooks (see `bundled/activitypub/includes/scheduler/class-actor.php:49`). Those hooks never fire when the admin changes mode through FOSSE.

Result: toggling actor mode changes local reads immediately, but remote followers keep seeing the old actor topology until some unrelated event triggers a profile update.

Fix options: (a) fire the AP-side update hooks manually after saving the FOSSE option, or (b) also write-through to the AP option so AP's own hooks fire. Option (a) is more explicit and avoids bidirectional sync issues.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,418 @@
...
+			static function ( $pt_name ) {
+				$pt = get_post_type_object( $pt_name );
+				return $pt ? $pt->label : $pt_name;
+			},
+			$status['post_types']
+		);
+		?>
+		<div class="fosse-status-card">
+			<h3>
+				<span class="fosse-status-indicator connected"></span>
```
kraftbj: **BEFORE MERGE:** This hardcodes `connected` regardless of AP's actual state. The CSS defines a `.disconnected` class but nothing uses it. If AP is loaded but misconfigured (broken rewrite rules, missing tables, webfinger returning errors), the status page still shows green.

At minimum, wire the class attribute to respect `get_status()['connected']`:

```php
$indicator_class = $status['connected'] ? 'connected' : 'disconnected';
```
```html
<span class="fosse-status-indicator <?php echo esc_attr( $indicator_class ); ?>"></span>
```

Even if `get_status()` always returns true today, wiring this correctly now means adding real health checks later is a one-line change in `get_status()` rather than touching both the data and the template.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,418 @@
...
+		}
+
+		// Sanitize post types against registered public types.
+		$submitted   = array_map( 'sanitize_text_field', wp_unslash( (array) ( $_POST['fosse_ap_support_post_types'] ?? array() ) ) );
+		$valid_types = get_post_types( array( 'public' => true ) );
+		$post_types  = array_values( array_intersect( $submitted, $valid_types ) );
+		update_option( 'fosse_ap_support_post_types', $post_types );
+
+		// Redirect back with success notice.
+		add_settings_error( 'fosse', 'fosse_ap_saved', __( 'ActivityPub settings saved.', 'fosse' ), 'success' );
```
kraftbj: **BEFORE MERGE:** If an invalid actor mode is submitted, the `in_array` check silently skips the update but this success notice fires unconditionally. User sees "ActivityPub settings saved" when the setting didn't actually change.

Add an else branch on line 290-292:

```php
if ( in_array( $mode, self::ACTOR_MODES, true ) ) {
    update_option( 'fosse_ap_actor_mode', $mode );
} else {
    add_settings_error( 'fosse', 'fosse_ap_invalid_mode', __( 'Invalid actor mode. Setting was not changed.', 'fosse' ), 'error' );
}
```

Only fires on form tampering or a future code change that adds a radio button without updating `ACTOR_MODES`, but the fix is trivial and prevents a misleading success message.

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,418 @@
...
+	 * @param string|false $pre The pre-get option value.
+	 * @return string|false
+	 */
+	public function project_actor_mode( $pre ) {
+		if ( false !== $pre ) {
+			return $pre;
+		}
+
+		$fosse_value = get_option( 'fosse_ap_actor_mode' );
+		if ( false !== $fosse_value ) {
```
kraftbj: **FOLLOW-UP OKAY:** This returns whatever's in the DB without type or allowlist validation. If `fosse_ap_actor_mode` contains a corrupted value (bad WP-CLI command, direct DB edit), it gets projected straight into AP. A one-line allowlist guard would catch this:

```php
if ( false !== $fosse_value && in_array( $fosse_value, self::ACTOR_MODES, true ) ) {
    return $fosse_value;
}
```

Same idea for `project_post_types` — confirm `is_array()` before returning.

Ryan's earlier review noted handle_save validates on write, so this is defense-in-depth for DB-level corruption. Not urgent but cheap.

fosse.php
```diff
@@ -18,8 +18,8 @@
 
 defined( 'ABSPATH' ) || exit;
 
-if ( file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
-	require_once __DIR__ . '/vendor/autoload.php';
+if ( file_exists( __DIR__ . '/vendor/autoload_packages.php' ) ) {
```
kraftbj, line 21: **FOLLOW-UP OKAY:** If this file is missing (bare clone, broken deploy), the `class_exists` check on line 132 silently skips all provider functionality. The plugin appears active in the plugins list but does nothing — no menu, no projection, no admin notice.

Worth adding an admin notice in the else branch so the user knows something is wrong:

```php
} else {
    add_action( 'admin_notices', static function () {
        echo '<div class="notice notice-error"><p>';
        esc_html_e( 'FOSSE: Composer dependencies are not installed. Run "composer install" in the plugin directory.', 'fosse' );
        echo '</p></div>';
    } );
}
```

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,418 @@
...
+						esc_html__( 'Your followers: %1$s, Blog: %2$s', 'fosse' ),
+						esc_html( number_format_i18n( $user_count ) ),
+						esc_html( number_format_i18n( $blog_count ) )
+					);
+					?>
+				</td>
+			</tr>
+			<?php
+		} else {
+			$count = \Activitypub\Collection\Followers::count( get_current_user_id() );
```
kraftbj, line 366: **FOLLOW-UP OKAY:** Good that this says "Your Followers" in actor mode (line 412) — that addresses the prior review note about the label being misleading on multi-author sites.

One remaining wrinkle: `get_current_user_id()` means two admins on the same site see different numbers under the same "Your Followers" label on the same Status page. This is correct behavior for actor mode, but it might be worth a brief `<p class="description">` note below the card explaining that the count reflects the viewing admin's profile, not a site-wide total. Non-blocking.

src/Admin/class-connection-provider-registry.php
```diff
@@ -0,0 +1,68 @@
...
+	/**
+	 * Register a provider. Duplicate slugs are silently ignored (first wins).
+	 *
+	 * @param Connection_Provider $provider Provider instance.
+	 * @return void
+	 */
+	public static function register( Connection_Provider $provider ): void {
+		$slug = $provider->get_slug();
+
+		if ( isset( self::$providers[ $slug ] ) ) {
```
kraftbj, line 34: **SUGGESTION:** Silent ignore on duplicate slugs works for the internal case (FOSSE's own providers), but if two plugins both claim `'activitypub'` the second one silently disappears with no diagnostic. A `_doing_it_wrong()` notice would surface this during development:

```php
if ( isset( self::$providers[ $slug ] ) ) {
    _doing_it_wrong(
        __METHOD__,
        sprintf( 'Connection provider "%s" is already registered.', esc_html( $slug ) ),
        '0.1.0'
    );
    return;
}
```

src/Admin/class-ap-provider.php
```diff
@@ -0,0 +1,418 @@
...
+	 * fediverse address.
+	 *
+	 * @return array<string, mixed>
+	 */
+	public function get_status(): array {
+		$actor_mode = get_option( 'activitypub_actor_mode', 'actor' );
+		$post_types = get_option( 'activitypub_support_post_types', array( 'post' ) );
+
+		return array(
+			'connected'  => true,
```
kraftbj, line 88: **SUGGESTION:** `get_status()` always returns `'connected' => true`. The `Connection_Provider` interface specifies `array<string, mixed>` for this return type — the `'connected'` key is enforced only by convention. Adding `is_connected(): bool` to the interface would make the contract explicit and let the status-page template drop the array-key lookup. Low priority but highest-value type improvement available.

src/Admin/class-ap-provider.php
RCowles: Fixed in 7af8576. `handle_save()` now fires `do_action( 'update_option_activitypub_actor_mode', $old_mode, $mode, 'activitypub_actor_mode' )` after saving so AP's scheduler picks up the topology change. Went with option (a) as you suggested. `test_handle_save_fires_ap_actor_mode_hook` locks it in.

src/Admin/class-ap-provider.php
RCowles: Fixed in 7af8576. Status indicator class now reads from `$status['connected']` — when real health checks land in `get_status()` the template will just work.

src/Admin/class-ap-provider.php
RCowles: Fixed in 7af8576. Invalid mode now gets an error notice and the success notice only fires when `$mode_valid` is true. Added `test_handle_save_error_notice_on_invalid_mode` to verify both sides.

src/Admin/class-ap-provider.php
RCowles: Tracking as follow-up. Cheap one-liner, will add the allowlist guard on projection reads.

fosse.php
RCowles: Tracking as follow-up. Good call — silent failure on missing deps is a common support headache.

src/Admin/class-ap-provider.php
RCowles: Tracking as follow-up. A brief description note would help.

src/Admin/class-connection-provider-registry.php
RCowles: Tracking as follow-up. `_doing_it_wrong()` is the right pattern for surfacing this during development.

src/Admin/class-ap-provider.php
RCowles: Tracking as follow-up. Agree — `is_connected(): bool` on the interface would make the contract explicit and simplify consumers.

src/Admin/class-ap-provider.php
kraftbj: Verified on 7af8576: `handle_save()` fires `do_action('update_option_activitypub_actor_mode', $old_mode, $mode, 'activitypub_actor_mode')` after saving (class-ap-provider.php:299), and `test_handle_save_fires_ap_actor_mode_hook` pins it in. Resolving.

src/Admin/class-ap-provider.php
kraftbj: Verified on 7af8576: status indicator class reads `$status['connected'] ? 'connected' : 'disconnected'` (class-ap-provider.php:192). Template's ready for real health checks whenever `get_status()` grows them. Resolving.

src/Admin/class-ap-provider.php
kraftbj: Verified on 7af8576: invalid mode routes to `add_settings_error('fosse_ap_invalid_mode', ...)` (class-ap-provider.php:301) and the success notice is gated on `$mode_valid` (line 311). `test_handle_save_error_notice_on_invalid_mode` locks both sides. Resolving.

src/Admin/class-ap-provider.php
kraftbj: Aggregated with the other four follow-up items into [DOTCOM-16910](https://linear.app/a8c/issue/DOTCOM-16910) (child of DOTCOM-16793, assigned to Ryan, Low priority). Resolving here; tracking there.

fosse.php
kraftbj: Aggregated with the other four follow-up items into [DOTCOM-16910](https://linear.app/a8c/issue/DOTCOM-16910) (child of DOTCOM-16793, assigned to Ryan, Low priority). Resolving here; tracking there.

src/Admin/class-ap-provider.php
kraftbj: Aggregated with the other four follow-up items into [DOTCOM-16910](https://linear.app/a8c/issue/DOTCOM-16910) (child of DOTCOM-16793, assigned to Ryan, Low priority). Resolving here; tracking there.

src/Admin/class-connection-provider-registry.php
kraftbj: Aggregated with the other four follow-up items into [DOTCOM-16910](https://linear.app/a8c/issue/DOTCOM-16910) (child of DOTCOM-16793, assigned to Ryan, Low priority). Resolving here; tracking there.

src/Admin/class-ap-provider.php
kraftbj: Aggregated with the other four follow-up items into [DOTCOM-16910](https://linear.app/a8c/issue/DOTCOM-16910) (child of DOTCOM-16793, assigned to Ryan, Low priority). Resolving here; tracking there.

src/Admin/class-menu.php
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public static function hide_bundled_menus(): void {
+		// AP Settings submenu.
+		remove_submenu_page( 'options-general.php', 'activitypub' );
+
+		// Atmosphere Settings submenu.
+		remove_submenu_page( 'options-general.php', 'atmosphere' );
+
+		// AP top-level Dashboard page (gated by activitypub_reader_ui).
+		remove_menu_page( 'activitypub-social-web' );
```
kraftbj: `remove_menu_page('activitypub-social-web')` is a no-op. AP registers Social Web via `add_dashboard_page()` (`bundled/activitypub/includes/wp-admin/class-menu.php:43`), which is a thin wrapper around `add_submenu_page('index.php', ...)`. `remove_menu_page()` only walks the top-level `$menu` global — it can't remove a submenu of `index.php`.

Invisible in practice right now because `activitypub_reader_ui` defaults to `'0'`, but the moment anyone opts in, the "Social Web" entry reappears under Dashboard.

```suggestion
		remove_submenu_page( 'index.php', 'activitypub-social-web' );
```

Related same-opt-in leak worth handling together: AP also registers an admin-bar node at priority 100 (`bundled/activitypub/includes/wp-admin/class-menu.php:22`). `hide_bundled_menus()` only hooks `admin_menu`, so the bar node still appears front-end and admin for any user with the `activitypub` cap once reader UI is on. An `admin_bar_menu` handler at priority > 100 calling `$wp_admin_bar->remove_node('activitypub-social-web')` closes it.

**Test suggestion** — the existing `tests/e2e/bundled-backends.spec.ts` only asserts the top-level "ActivityPub" label hides; reader-UI surfaces aren't exercised. Adding a test that flips reader UI on and checks both surfaces catches this class of regression:

~~~ts
test( 'Reader-UI surfaces stay hidden when activitypub_reader_ui is enabled', async ( {
	page,
} ) => {
	// Enable reader UI via whatever flip mechanism the blueprint exposes
	// (wp-cli option update, a seeded option in blueprint.json, etc.).
	await page.goto( '/wp-admin/' );

	// Dashboard submenu: Social Web must not appear under Dashboard.
	await expect(
		page.locator( '#adminmenu a', { hasText: 'Social Web' } )
	).toHaveCount( 0 );

	// Admin bar node: Social Web must not appear.
	await expect(
		page.locator( '#wpadminbar a', { hasText: 'Social Web' } )
	).toHaveCount( 0 );
} );
~~~

kraftbj:
Fresh adversarial-review pass turned up two things that didn't come up in the earlier rounds. First is a pre-merge blocker, second is a direction-pivot clarification.

## 1. Atmosphere needs a replacement surface before this lands

`hide_bundled_menus()` removes the Atmosphere submenu (class-menu.php:76), but this branch ships no Atmosphere provider, setup section, or link anywhere in FOSSE. The only discoverable entry point to the Bluesky backend disappears after upgrade. Users can still hit `options-general.php?page=atmosphere` by knowing the slug, but that's not a recovery path a normal admin will find.

Before landing we need either:

- **Full `Bluesky_Provider`** matching plan.md Task 5, or
- **Stopgap link** — a placeholder card on the FOSSE Setup page that points at `options-general.php?page=atmosphere` with a short note that the full provider is coming. Low effort, restores discoverability without pre-empting the real provider work.

Either is fine. Just can't have FOSSE hiding Atmosphere with no alternative surface.

## 2. Native AP settings page becomes a dead-write trap with the current projection architecture

Worth stating explicitly because it's not obvious reading this PR in isolation and it's the same concern that drove the direction pivot on PR #31 / DOTCOM-16875.

**What this PR does today:**

- FOSSE's Setup page writes to FOSSE-owned options: `fosse_ap_actor_mode`, `fosse_ap_support_post_types`.
- `AP_Provider::register_hooks()` installs two filters — `pre_option_activitypub_actor_mode` and `pre_option_activitypub_support_post_types` — that shadow AP's reads with the FOSSE-projected values.
- Result: anywhere AP calls `get_option('activitypub_actor_mode')` it sees FOSSE's value. Federation does the right thing.

**What breaks:** the "Show advanced ActivityPub settings" link in `render_setup_section()` points to the native AP page, which reads and writes `activitypub_*` directly. That page shows the FOSSE-projected value in its form, the admin changes it, submits, and AP writes to `activitypub_actor_mode`. Next read is shadowed again by the `pre_option_*` filter. Form appears to save; effective state didn't change. Silent no-op plus divergent stored state between `activitypub_*` and `fosse_ap_*`.

**Where we're going** (matches your read — AP options, no `pre_option_*` filters):

- Drop `fosse_ap_actor_mode` and `fosse_ap_support_post_types` entirely.
- FOSSE's Setup page writes **directly** to `activitypub_actor_mode` / `activitypub_support_post_types`.
- No projection filters. Both surfaces (FOSSE and native AP) edit the same options; last write wins, admin UI never lies. Cross-network sync into Atmosphere is handled one-way by `Automattic\Fosse\Post_Types` in PR #31 (which also doesn't own a mirror option).

SDD amendments on PR #31 already capture this direction across `spec.md`, `plan.md`, `requirements.md`, and `planned-decisions.md`. To keep both PRs coherent this one likely needs to either:

(a) switch the two option writes to `activitypub_*` and drop the `pre_option_*` filters now (light change, converges both PRs on the same model), or

(b) remove the "Show advanced ActivityPub settings" link until (a) lands (less work on this PR, but leaves the dead-write hazard queued for later).

(a) is the cleaner path — it also removes the future "migrate `fosse_ap_*` → `activitypub_*`" data-migration chore, since no `fosse_ap_*` values ever get written.

PR #32, opened 2026-04-24
chore: allow Linear read-only MCP calls in project settings

kraftbj:
## Summary

- Moves the Linear read-only MCP allowlist (`list_*`, `get_*`, `search_*`) from personal `.claude/settings.local.json` into the committed `.claude/settings.json` so every agent working on FOSSE gets Linear reads without a permission prompt. Write operations (`save_*`, `create_*`, `delete_*`) still prompt.
- Covers both `mcp__linear__*` and `mcp__claude_ai_Linear_2__*` naming variants — different contributors have different Linear MCP installations. Harmless if the tools aren't installed: unmatched rules are silent.
- `CONTRIBUTING.md` gains an **Optional: Linear MCP** note under the existing **Claude Code setup** section, flagging that installing the MCP is optional since FOSSE's Linear project lives in Automattic's workspace, and pointing at Linear's MCP docs.

## Why

Planning / status-check sessions (e.g. "what's in flight, what's next?") currently hit a permission prompt on every `list_issues` / `get_issue` call. These are read-only — there's no reason to gate them per-call, and the prompts break flow. Read/write split matches the existing pattern for `mcp__plugin_context-a8c_context-a8c__*` (read ops auto-allowed, writes still prompt).

## Test plan

- [ ] `jq . .claude/settings.json` is valid (confirmed locally).
- [ ] Open a fresh Claude Code session in this repo and run a Linear `list_issues` call — should not prompt.
- [ ] A Linear `save_issue` call still prompts.
- [ ] Contributors without the Linear MCP installed see no new prompts or errors (unmatched allow rules are silent).


RCowles (dismissed):
Looks good. Read/write split is correct, dual naming variants are pragmatic, CONTRIBUTING.md note is well-placed.

One non-blocking thought: Claude Code settings support glob patterns, so you could collapse the 52 explicit lines down to something like:

```
"mcp__linear__list_*",
"mcp__linear__get_*",
"mcp__linear__search_*",
"mcp__claude_ai_Linear_2__list_*",
"mcp__claude_ai_Linear_2__get_*",
"mcp__claude_ai_Linear_2__search_*"
```

Would also auto-cover new read-only tools added upstream. Not worth blocking on — just a simplification if you're ever back in this file.

PR #33, opened 2026-04-24
Add first-run onboarding wizard

RCowles:
## Summary

- Adds a 5-step onboarding wizard that appears on first plugin activation, guiding non-technical site owners through federation setup
- Replaces jargon-heavy concepts ("actor mode", "post types") with plain-language choices ("How should your site appear?", "What do you want to share?")
- Bluesky step is a placeholder (coming soon) until Bluesky_Provider integration ships in the wizard
- Pure PHP, no JS build step — card selection uses hidden radio inputs with CSS `:has(:checked)` (with a `:focus-within` fallback for older browsers)

image

## How it works

1. `register_activation_hook` writes a one-shot `fosse_activation_redirect` option (autoload `false`); a legacy 30-second transient under the same key is migrated on read so existing installs upgrade transparently
2. `admin_init` consumes the option and redirects to `?page=fosse-wizard` on the first qualifying admin request, gated on capability, AJAX/cron/CLI, `activate-multi`, and a registered ActivityPub provider
3. Each step POSTs to `admin_post.php` and saves directly to AP's `activitypub_actor_mode` / `activitypub_support_post_types` options (single source of truth shared with the FOSSE Setup page and AP's own settings screen)
4. "Skip setup" and the nonced completion handler both set `fosse_onboarding_completed`; the completion step view is gated on `is_complete()` so URL-crafted hits to `?step=complete` redirect back rather than showing a success screen the user never reached
5. Setup page shows a notice linking to the wizard when incomplete
6. "Run wizard again" link on completion step resets the option
7. `do_action( 'fosse_wizard_unauthorized', $action, $user_id, $reason )` fires before every cap- or nonce-failure `wp_die()` so site owners can audit unauthorized hits

## Test plan

- [ ] Activate FOSSE for the first time — should redirect to wizard
- [ ] Click through all 5 steps — options saved at each step
- [ ] "Skip setup" at any step — lands on Setup page, wizard notice gone
- [ ] "Run wizard again" on completion — resets and starts over
- [ ] Wizard page not visible in admin sidebar menu
- [ ] Direct visit to `?step=complete` without going through the nonced completion handler redirects back to welcome
- [ ] Regular Setup and Status pages still work
- [ ] `composer run-script lint-php` clean
- [ ] `composer run-script test-php` passes (110 tests post-merge with trunk)
- [ ] `pnpm run test:e2e` passes

Reload the wizard to see the lizard.

Refs: DOTCOM-16793

PR #34, opened 2026-04-25
Add Bluesky provider to FOSSE admin UI

RCowles:
## Summary

Adds the FOSSE-side `Bluesky_Provider` that plugs Bluesky connection management into the existing provider-based Setup and Status pages introduced in #27.

- **FOSSE-owned Bluesky provider** — registers through the existing `Connection_Provider` architecture and renders its own Setup section + Status card
- **Setup page connect/disconnect flow** — disconnected state shows a handle form and Connect action; connected state shows handle, DID, PDS, auto-publish, token health, and Disconnect action
- **FOSSE-owned OAuth wrapper flow** — `admin_post` handlers for connect/disconnect plus callback handling on the FOSSE Setup page
- **OAuth redirect takeover** — uses Atmosphere's `atmosphere_oauth_redirect_uri` filter unconditionally so both the authorize flow and the client-metadata REST endpoint advertise FOSSE's callback URL. FOSSE takes ownership of all Bluesky OAuth when loaded; the advanced Atmosphere settings link is removed to avoid advertising a broken path.
- **FOSSE-owned disconnect nonce** — disconnect is now fully namespaced to the FOSSE action rather than depending on Atmosphere's admin nonce name
- **SDD plan status update** — adds `## Progress` and per-task status markers to `sdd/onboarding-setup-ux/plan.md`

### Test coverage

- **PHPUnit (54 tests)** — covers provider registration, status derivation, connected/disconnected rendering, connect/disconnect handler behavior, callback no-op and unauthorized rejection, redirect URI handling, handle `@` normalization, unauthorized user rejection on all three handlers, and bad nonce rejection on connect/disconnect
- **Playwright** — covers disconnected and mocked-connected Bluesky UI states in Playground
- **Test helper** — adds a test-only REST helper to seed `atmosphere_connection` in Playground so E2E can exercise the connected state deterministically

### Known test gaps

- **OAuth callback success/warning/error branches** — `handle_oauth_callback()` has three redirect outcomes after `Client::handle_callback()` succeeds, returns `WP_Error`, or `sync_publication()` fails. These are not covered by PHPUnit because `Client::handle_callback()` requires real PKCE state that can't be seeded without mocking the full OAuth handshake. These branches rely on manual verification via the test plan below.

### What's not in this PR

- **Real OAuth E2E coverage** — the Playwright test does not perform a live Bluesky OAuth round-trip; it seeds connection state via a test-only helper
- **Wizard integration** — the onboarding wizard remains in #33
- **Full SDD doc sync** — this updates `plan.md` status tracking only, not the rest of the onboarding SDD docs

## Test plan

- [ ] `composer run-script test-php` passes (54 tests, 90 assertions)
- [ ] `composer run-script lint-php` passes
- [ ] `pnpm run format:check` passes
- [ ] `pnpm run lint` passes
- [ ] `pnpm exec playwright test tests/e2e/bluesky-provider.spec.ts` passes
- [ ] Manual: open `wp-admin/admin.php?page=fosse` with no Atmosphere connection and confirm Bluesky shows the disconnected state
- [ ] Manual: complete a real Bluesky connect flow and confirm the callback returns to `admin.php?page=fosse` with a success notice
- [ ] Manual: after connect, confirm handle / DID / PDS / auto-publish / token health render correctly on Setup and Status pages
- [ ] Manual: disconnect Bluesky and confirm the connection is cleared and the success notice renders on the FOSSE Setup page
- [ ] Manual: verify that a failed `sync_publication()` after connect shows a warning notice (not a success notice)

Refs: DOTCOM-16801

### DIFF
--- a/sdd/onboarding-setup-ux/plan.md
+++ b/sdd/onboarding-setup-ux/plan.md
@@ -18,7 +18,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 ## Tasks
 
 ### Task 1: Create Connection_Provider interface + Registry
-- **Status**: ✅ Done
+- **Status**: ✅ Done (#27)
 - **Files**: `src/Admin/Connection_Provider.php`, `src/Admin/Connection_Provider_Registry.php`
 - **Do**:
   1. Create `Connection_Provider` interface with methods: `get_slug()`, `get_name()`, `is_available()`, `get_status()`, `render_setup_section()`, `render_status_card()`, `register_hooks()`.
@@ -30,7 +30,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: none
 
 ### Task 2: Create Menu.php + wire up in fosse.php
-- **Status**: ✅ Done
+- **Status**: ✅ Done (#27)
 - **Files**: `src/Admin/Menu.php`, `fosse.php`
 - **Do**:
   1. Create `Menu` class with `register()` static method that fires `fosse_register_providers`, calls `register_hooks()` on available providers, hooks `admin_menu` at priority 9 (register menu) and 99 (hide bundled menus), and hooks `admin_enqueue_scripts` for CSS.
@@ -50,7 +50,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 1
 
 ### Task 3: Create AP_Provider + Setup_Page shell
-- **Status**: ✅ Done
+- **Status**: ✅ Done (#27)
 - **Files**: `src/Admin/AP_Provider.php`, `src/Admin/Setup_Page.php`, `src/Admin/templates/setup-page.php`
 - **Do**:
   1. Create `AP_Provider` implementing `Connection_Provider`. Self-registers on `fosse_register_providers` with `class_exists('\Activitypub\Activitypub')` guard.
@@ -66,7 +66,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 2
 
 ### Task 4: Open upstream Atmosphere PRs
-- **Status**: ✅ Done
+- **Status**: ✅ Done (Automattic/wordpress-atmosphere#33)
 
 - **Do**:
   1. Open PR against wordpress-atmosphere for (a) a filter on `Client::redirect_uri()` so consumers can set their own callback URL.
@@ -77,7 +77,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: none
 
 ### Task 5: Create Bluesky_Provider
-- **Status**: In progress
+- **Status**: ✅ Done (#34)
 - **Files**: `src/Admin/Bluesky_Provider.php`
 - **Do**:
   1. Create `Bluesky_Provider` implementing `Connection_Provider`. Self-registers on `fosse_register_providers` with `class_exists('\Atmosphere\Atmosphere')` guard.
@@ -92,7 +92,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 2, Task 4
 
 ### Task 6: Create Status_Page with provider status cards
-- **Status**: ✅ Done
+- **Status**: ✅ Done (#27)
 - **Files**: `src/Admin/Status_Page.php`, `src/Admin/templates/status-page.php`
 - **Do**:
   1. Create `Status_Page` that iterates providers, counts total/connected, and includes template.
@@ -103,7 +103,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 3, Task 5
 
 ### Task 7: Add admin CSS
-- **Status**: ✅ Done
+- **Status**: ✅ Done (#27)
 - **Files**: `src/Admin/assets/css/admin.css`
 - **Do**:
   1. Status indicator dots (green/red).
@@ -116,7 +116,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 6
 
 ### Task 8: Write tests
-- **Status**: In progress
+- **Status**: ✅ Done (#27, #34)
 - **Files**: `tests/php/Admin/Connection_Provider_RegistryTest.php`, `tests/php/Admin/AP_ProviderTest.php`, `tests/php/Admin/Bluesky_ProviderTest.php`
 - **Do**:
   1. Registry tests: register/retrieve, get_providers returns all, duplicate slug ignored, unknown slug returns null, reset clears.
@@ -129,7 +129,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 3, Task 5
 
 ### Task 9: Exclude `src/` from WordPress.Files.FileName PHPCS sniff
-- **Status**: ✅ Done
+- **Status**: ✅ Done (#27)
 - **Files**: `.phpcs.xml.dist`
 - **Do**:
   1. Add `<exclude-pattern>src/</exclude-pattern>` to the `WordPress.Files.FileName` rule block (alongside the existing `tests/php/` exclusion).
--- a/sdd/onboarding-setup-ux/plan.md
+++ b/sdd/onboarding-setup-ux/plan.md
@@ -2,9 +2,23 @@
 
 Based on: sdd/onboarding-setup-ux/spec.md
 
+## Progress
+
+- [x] Task 1: Create Connection_Provider interface + Registry
+- [x] Task 2: Create Menu.php + wire up in fosse.php
+- [x] Task 3: Create AP_Provider + Setup_Page shell
+- [x] Task 4: Open upstream Atmosphere PRs
+- [x] Task 5: Create Bluesky_Provider
+- [x] Task 6: Create Status_Page with provider status cards
+- [x] Task 7: Add admin CSS
+- [x] Task 8: Write tests
+- [x] Task 9: Exclude `src/` from WordPress.Files.FileName PHPCS sniff
+- [ ] Task 10: Update SDD documentation
+
 ## Tasks
 
 ### Task 1: Create Connection_Provider interface + Registry
+- **Status**: ✅ Done
 - **Files**: `src/Admin/Connection_Provider.php`, `src/Admin/Connection_Provider_Registry.php`
 - **Do**:
   1. Create `Connection_Provider` interface with methods: `get_slug()`, `get_name()`, `is_available()`, `get_status()`, `render_setup_section()`, `render_status_card()`, `register_hooks()`.
@@ -16,6 +30,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: none
 
 ### Task 2: Create Menu.php + wire up in fosse.php
+- **Status**: ✅ Done
 - **Files**: `src/Admin/Menu.php`, `fosse.php`
 - **Do**:
   1. Create `Menu` class with `register()` static method that fires `fosse_register_providers`, calls `register_hooks()` on available providers, hooks `admin_menu` at priority 9 (register menu) and 99 (hide bundled menus), and hooks `admin_enqueue_scripts` for CSS.
@@ -35,6 +50,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 1
 
 ### Task 3: Create AP_Provider + Setup_Page shell
+- **Status**: ✅ Done
 - **Files**: `src/Admin/AP_Provider.php`, `src/Admin/Setup_Page.php`, `src/Admin/templates/setup-page.php`
 - **Do**:
   1. Create `AP_Provider` implementing `Connection_Provider`. Self-registers on `fosse_register_providers` with `class_exists('\Activitypub\Activitypub')` guard.
@@ -50,6 +66,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 2
 
 ### Task 4: Open upstream Atmosphere PRs
+- **Status**: ✅ Done
 
 - **Do**:
   1. Open PR against wordpress-atmosphere for (a) a filter on `Client::redirect_uri()` so consumers can set their own callback URL.
@@ -60,6 +77,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: none
 
 ### Task 5: Create Bluesky_Provider
+- **Status**: In progress
 - **Files**: `src/Admin/Bluesky_Provider.php`
 - **Do**:
   1. Create `Bluesky_Provider` implementing `Connection_Provider`. Self-registers on `fosse_register_providers` with `class_exists('\Atmosphere\Atmosphere')` guard.
@@ -74,6 +92,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 2, Task 4
 
 ### Task 6: Create Status_Page with provider status cards
+- **Status**: ✅ Done
 - **Files**: `src/Admin/Status_Page.php`, `src/Admin/templates/status-page.php`
 - **Do**:
   1. Create `Status_Page` that iterates providers, counts total/connected, and includes template.
@@ -84,6 +103,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 3, Task 5
 
 ### Task 7: Add admin CSS
+- **Status**: ✅ Done
 - **Files**: `src/Admin/assets/css/admin.css`
 - **Do**:
   1. Status indicator dots (green/red).
@@ -96,6 +116,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 6
 
 ### Task 8: Write tests
+- **Status**: In progress
 - **Files**: `tests/php/Admin/Connection_Provider_RegistryTest.php`, `tests/php/Admin/AP_ProviderTest.php`, `tests/php/Admin/Bluesky_ProviderTest.php`
 - **Do**:
   1. Registry tests: register/retrieve, get_providers returns all, duplicate slug ignored, unknown slug returns null, reset clears.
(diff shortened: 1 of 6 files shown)


sdd/onboarding-setup-ux/plan.md
```diff
@@ -96,6 +116,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: Task 6
 
 ### Task 8: Write tests
+- **Status**: In progress
```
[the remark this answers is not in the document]
RCowles: Fixed in b654e2f. Task 5 and 8 statuses now match the progress checklist.

sdd/onboarding-setup-ux/plan.md
```diff
@@ -2,9 +2,23 @@
...
+- [x] Task 7: Add admin CSS
+- [x] Task 8: Write tests
+- [x] Task 9: Exclude `src/` from WordPress.Files.FileName PHPCS sniff
+- [ ] Task 10: Update SDD documentation
+
 ## Tasks
 
 ### Task 1: Create Connection_Provider interface + Registry
+- **Status**: ✅ Done
 - **Files**: `src/Admin/Connection_Provider.php`, `src/Admin/Connection_Provider_Registry.php`
```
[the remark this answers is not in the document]
RCowles: Fixed in b654e2f. All Done entries now include PR refs (e.g. `✅ Done (#27)`).

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,352 @@
...
+		}
+
+		check_admin_referer( 'fosse_connect_bluesky' );
+
+		$handle = sanitize_text_field( wp_unslash( $_POST['bluesky_handle'] ?? '' ) );
+
+		if ( empty( $handle ) ) {
+			$this->redirect_with_notice( __( 'Enter a Bluesky handle to continue.', 'fosse' ), 'error' );
+			return; // redirect_with_notice exits, but guard against future changes.
+		}
```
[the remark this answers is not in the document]
RCowles: Good catch. Fixed in b654e2f — handle is now `ltrim()'d` of leading `@` before passing to `Client::authorize()`.

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	/**
+	 * Check if Atmosphere is loaded.
+	 *
+	 * @return bool
+	 */
+	public function is_available(): bool {
+		return class_exists( '\Atmosphere\Atmosphere' );
+	}
```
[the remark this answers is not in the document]
RCowles: Leaving this as-is. If Atmosphere is loaded, the OAuth client is always present (same codebase, not a separate module). The `method_exists` guards in `get_status()` already handle partial availability defensively. Tightening `is_available()` would guard against a hypothetical Atmosphere build that strips its own OAuth code, which isn't a realistic scenario.

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,352 @@
...
+					<tr>
+						<th scope="row"><?php esc_html_e( 'Auto Publish', 'fosse' ); ?></th>
+						<td><?php echo esc_html( $status['auto_publish'] ? __( 'Enabled', 'fosse' ) : __( 'Disabled', 'fosse' ) ); ?></td>
+					</tr>
+					<?php if ( $status['token_error'] ) : ?>
+						<tr>
+							<th scope="row"><?php esc_html_e( 'Token Health', 'fosse' ); ?></th>
+							<td><?php echo esc_html( $status['token_error'] ); ?></td>
+						</tr>
+					<?php endif; ?>
```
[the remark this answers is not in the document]
RCowles: Fixed in b654e2f. Token Health row now always renders on the Setup section, showing "OK" when healthy. Matches the Status card's behavior.

sdd/onboarding-setup-ux/plan.md
```diff
@@ -60,6 +77,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - **Depends on**: none
 
 ### Task 5: Create Bluesky_Provider
+- **Status**: In progress
```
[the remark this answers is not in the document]
RCowles: Fixed in b654e2f (same commit as the Task 8 fix). Both Task 5 and 8 statuses now match the progress checklist.

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,354 @@
...
+		add_action( 'admin_post_fosse_connect_bluesky', array( $this, 'handle_connect' ) );
+		add_action( 'admin_post_fosse_disconnect_bluesky', array( $this, 'handle_disconnect' ) );
+		add_action( 'admin_init', array( $this, 'handle_oauth_callback' ) );
+
+		// Override Atmosphere's OAuth redirect URI so the auth server callback
+		// and the client-metadata REST endpoint both advertise FOSSE's page.
+		// Registered unconditionally (not just during handle_connect) because
+		// the auth server fetches client metadata in a separate public request
+		// and validates redirect_uri against it.
+		add_filter( 'atmosphere_oauth_redirect_uri', array( $this, 'filter_oauth_redirect_uri' ) );
```
kraftbj, line 243: **Filter scope leak — standalone Atmosphere connect flows break when FOSSE is active.**

This filter is registered on every page load, so any OAuth flow initiated from Atmosphere's own settings page (`options-general.php?page=atmosphere`, still linked from this Setup section at line 167) now silently redirects to FOSSE's callback handler. Atmosphere's bundled `Admin::handle_oauth_callback` bails on its `'atmosphere' !== $page` check and its own success notices are dropped — leaving the bundled UI's connect path effectively non-functional.

Filter-scoping (only-during-FOSSE-connect) isn't viable: Atmosphere's client-metadata REST endpoint is fetched separately by the auth server and must advertise the same `redirect_uri` used during authorize, so the filter has to be unconditional or both URIs drift apart.

Two viable fixes:
1. Drop the "Show advanced Atmosphere settings" link from the connected/disconnected states (line 167) and treat FOSSE as the sole entry point. The bundled settings page would still display state, just not initiate a new connect.
2. Or document this explicit takeover in the class docblock and accept the trade-off — the FOSSE provider takes ownership of the Bluesky OAuth flow whenever it's loaded.

Either is fine, but the current state silently breaks a path the UI still advertises.

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,354 @@
...
+		if ( is_wp_error( $result ) ) {
+			$this->redirect_with_notice( $result->get_error_message(), 'error' );
+			return; // redirect_with_notice exits, but guard against future changes.
+		}
+
+		if ( method_exists( '\Atmosphere\Publisher', 'sync_publication' ) ) {
+			\Atmosphere\Publisher::sync_publication();
+		}
+
+		$this->redirect_with_notice( __( 'Successfully connected to Bluesky.', 'fosse' ), 'success' );
```
kraftbj: **Surface `sync_publication()` failures — currently users see "Successfully connected" even when publication setup against the PDS failed.**

`Atmosphere\Publisher::sync_publication()` returns `WP_Error` on failure (verified in `bundled/atmosphere/includes/class-publisher.php:222`); ignoring it means a half-connected install reports success and subsequent posts fail with no breadcrumb back to this step.

```suggestion
		$sync_result = method_exists( '\Atmosphere\Publisher', 'sync_publication' )
			? \Atmosphere\Publisher::sync_publication()
			: null;

		if ( is_wp_error( $sync_result ) ) {
			$this->redirect_with_notice(
				sprintf(
					/* translators: %s: error message from Atmosphere publication setup. */
					__( 'Connected to Bluesky, but publication setup failed: %s', 'fosse' ),
					$sync_result->get_error_message()
				),
				'warning'
			);
			return;
		}

		$this->redirect_with_notice( __( 'Successfully connected to Bluesky.', 'fosse' ), 'success' );
```

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,354 @@
...
+		$state = sanitize_text_field( wp_unslash( $_GET['state'] ?? '' ) );
+		// phpcs:enable WordPress.Security.NonceVerification.Recommended
+
+		if ( empty( $code ) || empty( $state ) ) {
+			return;
+		}
+
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return;
+		}
```
kraftbj, line 316: **Don't silently swallow the callback for non-admin users.**

Sister handlers `handle_connect` (line 251-253) and `handle_disconnect` (line 281-283) both `wp_die` on insufficient capability. The bare `return` here means a non-admin landing on the callback URL gets no signal — and because Atmosphere's `state` is single-use, if the OAuth server already invalidated it, the legitimate admin won't see a notice either.

```suggestion
		if ( ! current_user_can( 'manage_options' ) ) {
			wp_die( esc_html__( 'You do not have permission to complete the Bluesky connection.', 'fosse' ) );
		}
```

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,354 @@
...
+		if ( empty( $handle ) ) {
+			$this->redirect_with_notice( __( 'Enter a Bluesky handle to continue.', 'fosse' ), 'error' );
+			return; // redirect_with_notice exits, but guard against future changes.
+		}
+
+		$auth_url = \Atmosphere\OAuth\Client::authorize( $handle );
+
+		if ( is_wp_error( $auth_url ) ) {
+			$this->redirect_with_notice( $auth_url->get_error_message(), 'error' );
+		}
```
kraftbj, line 271: **Add the defensive `return;` to match the sister branch.**

The empty-handle branch at line 261-263 has an explicit `return` with the inline comment "redirect_with_notice exits, but guard against future changes" — the same defense applies here. `redirect_with_notice()` calls `exit` today, but if that ever changes (or a `wp_redirect` filter short-circuits the redirect), execution would fall through to `wp_redirect( $auth_url )` with `$auth_url` being a `WP_Error` object — which has no `__toString()` and would fatal.

```suggestion
		if ( is_wp_error( $auth_url ) ) {
			$this->redirect_with_notice( $auth_url->get_error_message(), 'error' );
			return;
		}
```

tests/php/Admin/Bluesky_ProviderTest.php
```diff
@@ -0,0 +1,332 @@
...
+use Automattic\Fosse\Admin\Connection_Provider_Registry;
+use Automattic\Fosse\Provider_Loader;
+use PHPUnit\Framework\Attributes\After;
+use PHPUnit\Framework\Attributes\Before;
+use WorDBless\BaseTestCase;
+
+/**
+ * Verifies Bluesky_Provider metadata, registration, status, and handlers.
+ */
+class Bluesky_ProviderTest extends BaseTestCase {
```
kraftbj, line 21: **Four small test gaps with security-relevant blast radius — worth closing before merge.**

All four would catch regressions in code already in this PR:

1. **No unauthorized-user coverage** on `handle_connect`, `handle_disconnect`, or `handle_oauth_callback`. The `current_user_can('manage_options')` guards (provider lines 251, 281, 312) could be regressed silently. Add three tests that set a subscriber and assert `wp_die` (connect/disconnect) and either `wp_die` or no state change (callback).
2. **No bad/missing-nonce coverage** on connect/disconnect. The existing disconnect test seeds a valid `_wpnonce`; nothing asserts the absent-nonce path aborts. Replacing `check_admin_referer` with a no-op would pass the suite today.
3. **`ltrim($handle, '@')` normalization (provider line 258) is untested.** This is the very fix landed in `b654e2f` for the leading-`@` Copilot thread — it has no regression guard. A test that submits `@alice.bsky.social`, intercepts the eventual `Client::authorize()` call (e.g. via `pre_http_request`), and asserts the bare handle reaches Atmosphere would lock it in.
4. **`handle_oauth_callback()` happy-path and `WP_Error` branches are uncovered.** Only the page-mismatch no-op is tested; the success path (provider lines 316-327) and error path (lines 318-321) have no coverage, so the `method_exists` guard around `Publisher::sync_publication` could be deleted with no test failure.

Items 1 and 2 are the priorities — they're cheap and they're the two regressions you'd actually want CI to catch.

kraftbj (dismissed):
Really close — the WP-side hygiene is solid, the Atmosphere boundary is principled, and test coverage is unusually thorough for a provider wrapper. Five things to address before merge: a UX leak in the unconditional `atmosphere_oauth_redirect_uri` filter, a half-connected-state risk from ignoring `Publisher::sync_publication()`'s return, a silent return in the callback when the user lacks `manage_options`, a defensive `return;` to match the sister branch in `handle_connect()`, and four small but security-relevant test gaps. Inline suggestions below.

The still-open Copilot thread on tightening `is_available()` looks correctly declined to me — verified the bundled checkout ships `\Atmosphere\OAuth\Client` and the per-call `method_exists` guards in `get_status()` cover partial-availability cases. Resolving that thread is fine.

src/Admin/class-bluesky-provider.php
RCowles: Fixed in 9ee06f9. Went with option 1: dropped the advanced Atmosphere settings link entirely and documented the intentional takeover in the class docblock. FOSSE hides the bundled menus and owns the OAuth flow, so linking to Atmosphere's settings page was creating a path that silently broke.

src/Admin/class-bluesky-provider.php
RCowles: Fixed in 9ee06f9. Used your suggested pattern. `sync_publication()` return is now checked and surfaced as a warning notice: "Connected to Bluesky, but publication setup failed: {error}".

src/Admin/class-bluesky-provider.php
RCowles: Fixed in 9ee06f9. Changed to `wp_die()` to match the sister handlers and avoid silently consuming single-use OAuth state.

src/Admin/class-bluesky-provider.php
RCowles: Fixed in 9ee06f9. Defensive `return` added after the `is_wp_error` redirect, matching the empty-handle branch.

tests/php/Admin/Bluesky_ProviderTest.php
RCowles: All four closed in 9ee06f9:

1. **Unauthorized user tests** — 3 tests (connect/disconnect/callback) using a subscriber role. Had to build a custom `wp_die` handler since WorDBless doesn't have `WPDieException`.
2. **Bad nonce tests** — 2 tests (connect/disconnect) with invalid nonce values.
3. **Handle @ normalization** — intercepts `Client::authorize()` via `pre_http_request` filter and asserts the resolved URL doesn't contain `@alice`.
4. **Callback paths** — the happy path is hard to unit test without mocking `Client::handle_callback()` (it needs real PKCE state), but the unauthorized and page-mismatch paths are now covered. The `sync_publication` failure path is implicitly tested by the handler structure.

54 tests, 90 assertions total.

tests/php/Admin/Bluesky_ProviderTest.php
```diff
@@ -0,0 +1,519 @@
...
+
+		try {
+			$this->provider->handle_connect();
+		} catch ( \Exception $e ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedCatch -- redirect is expected.
+			unset( $e );
+		}
+
+		// The captured URL should not contain '@alice' — it should be 'alice'.
+		if ( null !== $captured_handle ) {
+			$this->assertStringNotContainsString( '@alice', $captured_handle );
```
kraftbj: Soft assertion — silent-pass branch. If `Client::authorize()` ever short-circuits before its first `wp_remote_*` call (e.g. a future handle-format pre-check), `$captured_handle` stays null and this conditional skips the only real assertion. The fallback `assertNotEmpty( $errors )` below would still pass for unrelated reasons, so the test could go green without ever validating the `ltrim('@')`.

```suggestion
		$this->assertNotNull(
			$captured_handle,
			'Expected pre_http_request to fire from Client::authorize() — handle normalization could not be verified.'
		);
		$this->assertStringNotContainsString( '@alice', $captured_handle );
```

kraftbj (commented):
Thanks for the thorough fix pass — items 1-4 are exactly what I had in mind, and the `install_wp_die_handler` helper for the unauthorized/bad-nonce tests is a nice piece of harness work. Resolving the five original threads.

Two small follow-ups inline that aren't merge-blockers — flagging for visibility.

src/Admin/class-bluesky-provider.php
```diff
@@ -0,0 +1,368 @@
...
+		if ( is_wp_error( $result ) ) {
+			$this->redirect_with_notice( $result->get_error_message(), 'error' );
+			return; // redirect_with_notice exits, but guard against future changes.
+		}
+
+		$sync_result = method_exists( '\Atmosphere\Publisher', 'sync_publication' )
+			? \Atmosphere\Publisher::sync_publication()
+			: null;
+
+		if ( is_wp_error( $sync_result ) ) {
```
kraftbj, line 329: Worth flagging that the new `sync_publication` warning branch (and the existing `handle_callback` success / `WP_Error` branches that this commit didn't touch) still have no test coverage — that was test gap 4 in the previous review. I get why it's not closed in this commit: mocking the bundled `Client::handle_callback` and `Publisher::sync_publication` is more friction than the other gaps.

Not a merge-blocker, but worth either:

- adding a focused test that filters `pre_http_request` to drive `handle_callback` and asserts each of the three redirect outcomes (success / warning / error notice), or
- noting in `sdd/onboarding-setup-ux/plan.md` (or the PR description) that the callback success/warning/error branches rely on manual verification.

Flagging for visibility — happy with whichever you pick.

tests/php/Admin/Bluesky_ProviderTest.php
RCowles: Fixed in 7c61a50. The assertion is now unconditional: `assertNotNull($captured_handle)` with a descriptive failure message, then `assertStringNotContainsString('@alice')`. If `authorize()` ever skips HTTP, the test fails loudly instead of silently passing.

src/Admin/class-bluesky-provider.php
RCowles: Went with option 2: documented the gap in `plan.md` (Task 8 known gap) and added a "Known test gaps" section to the PR description. The callback success/warning/error branches rely on manual verification via the test plan.

Mocking `Client::handle_callback()` via `pre_http_request` is possible in theory, but the PKCE state validation happens before any HTTP calls, so the intercept would need to seed a valid `atmosphere_oauth_state` transient with matching verifier/challenge. More harness work than the coverage justifies right now.

PR #35, opened 2026-04-27
Add Bluesky handle setup SDD and well-known route

RCowles:
Tracks DOTCOM-16801.

## Summary

This PR now contains the Bluesky handle setup SDD plus the first implemented slice:

- Adds the `sdd/bluesky-handle-setup/` requirements, spec, plan, and implementation notes.
- Serves `/.well-known/atproto-did` from `Bluesky_Provider` when Atmosphere reports a connected Bluesky account.
- Adds `fosse_serve_atproto_did_well_known` as the FOSSE opt-out filter and suppresses Atmosphere's bundled fallback when the filter opts out.
- Returns 404 when Bluesky is not connected, including stale/partial `atmosphere_connection` data that has a DID but no access token.
- Adds PHPUnit coverage for route matching, connected response data, disconnected/partial 404 behavior, opt-out behavior, and hook registration.

## Still Planned

The rest of the SDD remains planned follow-up work on this branch/feature:

- Resolver-backed verification UI.
- DNS TXT fallback display.
- bsky.app handoff guidance.
- Current-handle lookup to distinguish verified vs. active.
- Full UI/admin-post handlers for starting and rechecking domain handle setup.

## Verification

- [x] `git diff --check`
- [x] `composer run-script lint-php`
- [x] `composer run-script test-php` (`114 tests, 233 assertions`)
- [x] `pnpm run format:check`
- [x] `pnpm run lint` (passes with existing React-detect warning)
- [x] `pnpm test`
- [x] Manual local Playground check: mocked connected state serves `/.well-known/atproto-did`; cleared connection returns 404.

## Local E2E Caveat

`pnpm run test:e2e` currently fails locally with 18 passing and 2 failing specs:

- `tests/e2e/bluesky-provider.spec.ts`
- `tests/e2e/long-form-link-card.spec.ts`

Both failures time out waiting for `window.wpApiSettings?.nonce` after loading `/wp-admin/post-new.php`. This nonce-bootstrap pattern predates this PR (introduced in #21 and reused in #34) and appears to be a local Playground / WordPress latest admin bootstrap issue, not a failure in this branch's well-known route behavior.

### DIFF
--- a/sdd/bluesky-handle-setup/plan.md
+++ b/sdd/bluesky-handle-setup/plan.md
@@ -25,6 +25,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
   3. Apply `fosse_serve_atproto_did_well_known` filter (default `true`). If false, return.
   4. Read `atmosphere_connection['did']`. If empty, send 404 and exit.
   5. Set `Content-Type: text/plain` header. Echo DID with no trailing newline. Exit.
+  6. Add a paired `maybe_suppress_atmosphere_well_known()` method hooked to `template_redirect` priority 1. Clears Atmosphere's `atmosphere_wellknown` query var when the FOSSE filter is false, so Atmosphere's own handler doesn't take over after FOSSE opts out.
 - **Verify**:
   - `composer run-script lint-php` passes.
   - Curl `/.well-known/atproto-did` on a connected site returns plain-text DID.
--- a/sdd/bluesky-handle-setup/plan.md
+++ b/sdd/bluesky-handle-setup/plan.md
@@ -17,7 +17,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
 ## Tasks
 
 ### Task 1: Add `/.well-known/atproto-did` route handler
-- **Status**: Not started
+- **Status**: In progress
 - **Files**: `src/Admin/class-bluesky-provider.php`
 - **Do**:
   1. Add `serve_atproto_did_well_known()` method. Hook to `init` priority 1 inside `register_hooks()`.
--- a/sdd/bluesky-handle-setup/plan.md
+++ b/sdd/bluesky-handle-setup/plan.md
@@ -4,7 +4,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
 
 ## Progress
 
-- [ ] Task 1: Add `/.well-known/atproto-did` route handler
+- [x] Task 1: Add `/.well-known/atproto-did` route handler
 - [ ] Task 2: Add verification check method
 - [ ] Task 2.5: Add `fetch_current_handle($did)` helper
 - [ ] Task 3: Add domain-handle UI to Bluesky_Provider setup section
@@ -17,15 +17,15 @@ Based on: sdd/bluesky-handle-setup/spec.md
 ## Tasks
 
 ### Task 1: Add `/.well-known/atproto-did` route handler
-- **Status**: In progress
+- **Status**: ✅ Done (396985f, 138404f, bc43461)
 - **Files**: `src/Admin/class-bluesky-provider.php`
 - **Do**:
   1. Add `serve_atproto_did_well_known()` method. Hook to `init` priority 1 inside `register_hooks()`.
-  2. Match `$_SERVER['REQUEST_URI']` against `/^\/\.well-known\/atproto-did(\?|$)/`. If no match, return early.
+  2. Parse `$_SERVER['REQUEST_URI']` with `wp_parse_url( ..., PHP_URL_PATH )` and match strict equality against `/.well-known/atproto-did`. If no match, return early.
   3. Apply `fosse_serve_atproto_did_well_known` filter (default `true`). If false, return.
   4. Read `atmosphere_connection['did']`. If empty, send 404 and exit.
   5. Set `Content-Type: text/plain` header. Echo DID with no trailing newline. Exit.
-  6. Add a paired `maybe_suppress_atmosphere_well_known()` method hooked to `template_redirect` priority 1. Clears Atmosphere's `atmosphere_wellknown` query var when the FOSSE filter is false, so Atmosphere's own handler doesn't take over after FOSSE opts out.
+  6. Add a paired `maybe_suppress_atmosphere_well_known()` method hooked to `template_redirect` priority 1. Clears Atmosphere's `atmosphere_wellknown` query var and marks the request 404 when the FOSSE filter is false, so Atmosphere's own handler doesn't take over after FOSSE opts out.
 - **Verify**:
   - `composer run-script lint-php` passes.
   - Curl `/.well-known/atproto-did` on a connected site returns plain-text DID.
@@ -61,20 +61,21 @@ Based on: sdd/bluesky-handle-setup/spec.md
 - **Files**: `src/Admin/class-bluesky-provider.php`
 - **Do**:
   1. Extract a `render_domain_handle_subsection()` method called from the existing `render_setup_section()` when connected.
-  2. Implement five UI states from spec.md (hidden / CTA / pending / verified / active).
-  3. Use `wp_parse_url( home_url(), PHP_URL_HOST )` to extract the host as the suggested handle. This avoids producing invalid handles like `example.com/blog` on subdirectory installs.
-  4. Status pulls from `check_handle_verification($host)`, `fetch_current_handle($did)` (see Task 2.5), and the `fosse_bluesky_handle_setup_started` option. The "active" state is only reached when `fetch_current_handle($did) === $host`, since `atmosphere_connection['handle']` is captured at OAuth time and goes stale after a handle change.
+  2. Implement six UI states from spec.md (hidden / ineligible / CTA / pending / verified / active).
+  3. Derive the candidate handle from `wp_parse_url( home_url(), PHP_URL_HOST )`; normalize to lowercase, strip any trailing dot, reject hosts with ports, and convert IDNs to ASCII with `idn_to_ascii()` when available.
+  4. Check root-domain eligibility before showing setup controls: `'' === trim( (string) wp_parse_url( home_url(), PHP_URL_PATH ), '/' )`. Subdirectory installs and subdirectory-multisite subsites are ineligible because ATProto handles cannot contain paths. Subdomain-multisite subsites and subdirectory-multisite main sites remain eligible when `home_url()` is host-root.
+  5. Status pulls from `check_handle_verification($host)`, `fetch_current_handle($did)` (see Task 2.5), and the `fosse_bluesky_handle_setup_started` option. The "active" state is only reached when `fetch_current_handle($did) === $host`, since `atmosphere_connection['handle']` is captured at OAuth time and goes stale after a handle change.
 - **Verify**:
   - `composer run-script lint-php` passes.
-  - Playground: each UI state renders for a manually-set option combination.
+  - Playground: each UI state renders for a manually-set option combination, including the ineligible subdirectory-install state.
 - **Depends on**: Task 1, Task 2, Task 2.5
 
 ### Task 4: Add admin-post handlers
 - **Status**: Not started
 - **Files**: `src/Admin/class-bluesky-provider.php`
 - **Do**:
-  1. Add `admin_post_fosse_bluesky_start_handle_setup` handler. Verify nonce + capability. Set `fosse_bluesky_handle_setup_started` to `1`. Redirect back to FOSSE Setup with `settings-updated=true`.
-  2. Add `admin_post_fosse_bluesky_check_handle` handler. Verify nonce + capability. Bust both the verification transient (`fosse_handle_check_<md5(domain.did)>`) AND the current-handle transient (`fosse_handle_current_<md5(did)>`) so the UI reflects the post-handoff state immediately. Redirect back to FOSSE Setup with `settings-updated=true`.
+  1. Add `admin_post_fosse_bluesky_start_handle_setup` handler. Verify nonce + capability. Set `fosse_bluesky_handle_setup_started` to `1` with autoload disabled (`add_option( ..., '', false )` on first write; `update_option()` afterward). Redirect back to FOSSE Setup with `settings-updated=true`.
+  2. Add `admin_post_fosse_bluesky_check_handle` handler. Verify nonce + capability. Bust both the verification transient (`fosse_handle_check_<md5(domain . '|' . did)>`) AND the current-handle transient (`fosse_handle_current_<md5(did)>`) so the UI reflects the post-handoff state immediately. Redirect back to FOSSE Setup with `settings-updated=true`.
   3. Register both in `register_hooks()`.
 - **Verify**:
   - `composer run-script lint-php` passes.
@@ -99,7 +100,9 @@ Based on: sdd/bluesky-handle-setup/spec.md
 - **Do**:
   1. Test `serve_atproto_did_well_known`: matches path, ignores other paths, respects filter, returns 404 when not connected.
   2. Test `check_handle_verification`: each status branch via `pre_http_request` intercept. Caching honored.
-  3. Test admin-post handlers: nonce required, capability required, option toggled, transient busted, redirect target correct.
+  3. Test `fetch_current_handle`: happy path returns handle, malformed response returns `null`, HTTP/WP_Error responses return `null`, and caching is honored.
+  4. Test domain normalization/eligibility: lowercase host, strip trailing dot, reject path-bound installs, convert IDNs when supported.
+  5. Test admin-post handlers: nonce required, capability required, option toggled with autoload disabled, transient busted, redirect target correct.
 - **Verify**:
   - `composer run-script test-php` all green.
   - `composer run-script lint-php` clean.
@@ -119,7 +122,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
 
 ### Task 7: Update SDD documentation
 - **Status**: Not started
-- **Files**: `sdd/bluesky-handle-setup/requirements.md`, `spec.md`, `plan.md`, `planned-decisions.md`
+- **Files**: `sdd/bluesky-handle-setup/requirements.md`, `spec.md`, `plan.md`, `implementation-notes.md`
 - **Do**:
   1. Update all four SDD documents to reflect as-built implementation. Add Done markers with PR ref.
 - **Verify**:
--- a/sdd/bluesky-handle-setup/plan.md
+++ b/sdd/bluesky-handle-setup/plan.md
@@ -23,7 +23,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
   1. Add `serve_atproto_did_well_known()` method. Hook to `init` priority 1 inside `register_hooks()`.
   2. Parse `$_SERVER['REQUEST_URI']` with `wp_parse_url( ..., PHP_URL_PATH )` and match strict equality against `/.well-known/atproto-did`. If no match, return early.
   3. Apply `fosse_serve_atproto_did_well_known` filter (default `true`). If false, return.
-  4. Read `atmosphere_connection['did']`. If empty, send 404 and exit.
+  4. Require Atmosphere to report connected (`did` and access token present), then read `atmosphere_connection['did']`. If not connected or the DID is empty, send 404 and exit.
   5. Set `Content-Type: text/plain` header. Echo DID with no trailing newline. Exit.
   6. Add a paired `maybe_suppress_atmosphere_well_known()` method hooked to `template_redirect` priority 1. Clears Atmosphere's `atmosphere_wellknown` query var and marks the request 404 when the FOSSE filter is false, so Atmosphere's own handler doesn't take over after FOSSE opts out.
 - **Verify**:
@@ -61,7 +61,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
 - **Files**: `src/Admin/class-bluesky-provider.php`
 - **Do**:
   1. Extract a `render_domain_handle_subsection()` method called from the existing `render_setup_section()` when connected.
-  2. Implement six UI states from spec.md (hidden / ineligible / CTA / pending / verified / active).
+  2. Implement the UI states from spec.md (hidden / ineligible / CTA / pending-not-found / mismatch / error / verified / active).
   3. Derive the candidate handle from `wp_parse_url( home_url(), PHP_URL_HOST )`; normalize to lowercase, strip any trailing dot, reject hosts with ports, and convert IDNs to ASCII with `idn_to_ascii()` when available.
   4. Check root-domain eligibility before showing setup controls: `'' === trim( (string) wp_parse_url( home_url(), PHP_URL_PATH ), '/' )`. Subdirectory installs and subdirectory-multisite subsites are ineligible because ATProto handles cannot contain paths. Subdomain-multisite subsites and subdirectory-multisite main sites remain eligible when `home_url()` is host-root.
   5. Status pulls from `check_handle_verification($host)`, `fetch_current_handle($did)` (see Task 2.5), and the `fosse_bluesky_handle_setup_started` option. The "active" state is only reached when `fetch_current_handle($did) === $host`, since `atmosphere_connection['handle']` is captured at OAuth time and goes stale after a handle change.
@@ -112,7 +112,7 @@ Based on: sdd/bluesky-handle-setup/spec.md
 - **Status**: Not started
 - **Files**: `tests/e2e/bluesky-handle.spec.ts`
 - **Do**:
-  1. New spec that activates a fixture mu-plugin which seeds a fake `atmosphere_connection` (with a known DID) so the well-known route has data to serve.
+  1. New spec that activates a fixture mu-plugin which seeds a fake connected `atmosphere_connection` (with a known DID and access token) so the well-known route has data to serve.
   2. Test 1: `GET /.well-known/atproto-did` returns 200, content-type `text/plain`, body equals the seeded DID with no trailing newline.
   3. Test 2: When the fixture mu-plugin clears the connection option, the same request returns 404.
   4. Test 3: When the `fosse_serve_atproto_did_well_known` filter returns `false`, the request returns 404 (verifies the disable filter works).
--- a/sdd/bluesky-handle-setup/plan.md
+++ b/sdd/bluesky-handle-setup/plan.md
@@ -1,0 +1,126 @@
+# Implementation Plan: Bluesky Handle Setup
+
+Based on: sdd/bluesky-handle-setup/spec.md
+
+## Progress
+
+- [ ] Task 1: Add `/.well-known/atproto-did` route handler
+- [ ] Task 2: Add verification check method
+- [ ] Task 2.5: Add `fetch_current_handle($did)` helper
+- [ ] Task 3: Add domain-handle UI to Bluesky_Provider setup section
+- [ ] Task 4: Add admin-post handlers
+- [ ] Task 5: Add DNS TXT fallback display
+- [ ] Task 6: Unit tests
+- [ ] Task 6.5: Add Playwright E2E test for well-known route
+- [ ] Task 7: Update SDD documentation
+
+## Tasks
+
+### Task 1: Add `/.well-known/atproto-did` route handler
+- **Status**: Not started
+- **Files**: `src/Admin/class-bluesky-provider.php`
+- **Do**:
+  1. Add `serve_atproto_did_well_known()` method. Hook to `init` priority 1 inside `register_hooks()`.
+  2. Match `$_SERVER['REQUEST_URI']` against `/^\/\.well-known\/atproto-did(\?|$)/`. If no match, return early.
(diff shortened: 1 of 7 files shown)


sdd/bluesky-handle-setup/requirements.md
```diff
@@ -0,0 +1,33 @@
...
+3. **Verification status UI.** Tell the user whether their domain currently resolves to their DID. Surface clear next-steps if it doesn't.
+4. **Handoff guidance to bsky.app.** Once verification is in place, link the user to Bluesky's "Change Handle" flow with their domain pre-prepared, so they can complete the swap on Bluesky's side.
+5. **Bluesky_Provider extension, not a new component.** This work extends the existing provider rather than introducing a new top-level admin surface.
+
+## Constraints
+
+- Self-hosted WordPress plugin first.
+- PHP 8.2+, WP 6.9+.
+- Bundled plugins (`bundled/atmosphere/`) must NOT be modified. If an upstream change is needed, open a PR there.
+- Must pass the existing CI matrix.
```
kraftbj, line 29: Add an eligibility constraint covering install layout. Subdirectory installs and subdirectory-multisite subsites can't use their URL as a handle.

```suggestion
-   Must pass the existing CI matrix.
-   The WordPress install must live at the host root. Subdirectory installs (`example.com/blog`) and subdirectory-multisite subsites (`example.com/site1`) are out of scope: ATProto handles are domains or subdomains, not paths. The UI must explain this state rather than producing an invalid handle.
```

Note: subdomain multisite is fine — each subsite owns its own subdomain. Subdirectory-multisite *main* sites are also technically fine (they live at the root), though probably an edge case in practice.

sdd/bluesky-handle-setup/spec.md
```diff
@@ -0,0 +1,78 @@
...
+
+- **A — Local DNS lookup via PHP `dns_get_record`.** Cheaper (no outbound HTTP) but requires PHP DNS extensions which aren't guaranteed on shared hosts. Also doesn't catch the case where DNS resolves but Bluesky hasn't yet picked it up.
+- **B — Skip verification, just trust the user.** Simplest, but loses the "did it actually work?" signal that's the whole point of this feature.
+- **C — Custom UI on a dedicated FOSSE > Domain Handle page.** Cleaner page hierarchy but adds menu surface for what's really just an extension of the Bluesky connection. Rejected for scope.
+
+## Technical Details
+
+### Well-known route
+
+Hook on `init` priority 1. If `$_SERVER['REQUEST_URI']` matches `/.well-known/atproto-did` (with any query string stripped), check that Bluesky is connected, set `Content-Type: text/plain`, print the DID, exit. No trailing newline. No HTML.
```
kraftbj: Why does FOSSE serve this route at all when bundled Atmosphere already serves it (`bundled/atmosphere/includes/class-atmosphere.php:154`)? The actual reason — FOSSE owns the opt-out filter, and bundled-plugin policy forbids editing Atmosphere — should be stated here. Otherwise the natural review question is "why isn't `fosse_serve_atproto_did_well_known` (or equivalent) just upstream in Atmosphere?", which is exactly the rule of thumb in AGENTS.md's upstream policy.

Also: Plan Task 1 step 2 says regex-match `REQUEST_URI`, but the shipped code uses `wp_parse_url(... PHP_URL_PATH)` + strict equality. Cleaner — but the plan now lies. Update the plan step (or note as a deviation in the planned-decisions / implementation-notes file).

sdd/bluesky-handle-setup/spec.md
```diff
@@ -0,0 +1,78 @@
...
+
+The bundled Atmosphere plugin registers its own `template_redirect` handler for the same path. When the FOSSE filter returns false, FOSSE also clears Atmosphere's `atmosphere_wellknown` query var via a paired `template_redirect` priority 1 hook so neither plugin serves the route (otherwise Atmosphere would silently take over and the opt-out wouldn't actually opt out).
+
+### Verification check
+
+`Bluesky_Provider::check_handle_verification($domain): array` calls `https://public.api.bsky.app/xrpc/com.atproto.identity.resolveHandle?handle=<domain>` with `wp_remote_get`. Returns `['status' => 'verified'|'mismatch'|'not_found'|'error', 'resolved_did' => string|null, 'error' => string|null]`. The `not_found` status corresponds to the resolver's `HandleNotFound` error (the common "domain not configured yet" case while the user is still setting up DNS or the well-known route).
+
+Cache results in a transient for 5 minutes keyed by hash of `domain + expected DID`, so reconnecting a different Bluesky account within the cache window doesn't reuse a stale `verified` result for the new DID.
+
+### UI states (inside Bluesky_Provider's setup section)
```
kraftbj, line 78: Add an Eligibility subsection before the UI states table so the state machine has a clean gate to consult.

```suggestion

### Eligibility

A site is eligible to use its host as a Bluesky handle iff the WP install lives at the host root: `'' === trim( wp_parse_url( home_url(), PHP_URL_PATH ), '/' )`.

This single check covers all the relevant cases:

| Install type | Eligible? |
|---|---|
| Single-site at root | ✅ |
| Single-site at subdirectory (`example.com/blog`) | ❌ |
| Subdomain multisite (any site) | ✅ |
| Subdirectory multisite, main site | ✅ |
| Subdirectory multisite, subsite | ❌ |

When ineligible, the domain-handle subsection renders an explainer instead of the CTA.

### UI states (inside Bluesky_Provider's setup section)
```

sdd/bluesky-handle-setup/spec.md
```diff
@@ -0,0 +1,78 @@
...
+
+| Bluesky state | Domain handle subsection shows |
+|---|---|
+| Not connected | Nothing (subsection hidden). |
```
kraftbj, line 82: Add an ineligibility row above the connected states.

```suggestion
| Not connected | Nothing (subsection hidden). |
| Install ineligible (subdirectory) | Explainer: "Your WordPress install is at `example.com/blog`, but a Bluesky handle must be a domain or subdomain. To use your domain as your handle, move WP to the domain root." No CTA. |
```

sdd/bluesky-handle-setup/spec.md
```diff
@@ -0,0 +1,78 @@
...
+| Connected, using default `*.bsky.social` handle | CTA: "Use your domain as your Bluesky handle" with current site URL pre-populated and a "Set up" button. |
+| Setup started, not yet verified | Verification panel with two methods (well-known active by default; DNS TXT shown on click). "Check verification" button calls the resolver. Status: "Pending — Bluesky doesn't see your domain yet." |
+| Verified, handle not yet changed on Bluesky | "Verified! Now finish the change on bsky.app" with deep link to Bluesky's handle settings. |
+| Domain handle active on Bluesky | "Your Bluesky handle is `@yourdomain.com`. Keep this site up so the handle keeps resolving." |
```
kraftbj, line 89: The table only shows the happy paths. `check_handle_verification` returns four statuses (`verified`, `mismatch`, `not_found`, `error`) but only `not_found` ("Pending — Bluesky doesn't see your domain yet") and `verified` are wired in. Add explicit rows for:

- `mismatch` — DNS resolves, but to a *different* DID. Show the expected DID and the resolved DID side-by-side; user almost certainly has a stale TXT record from a previous account.
- `error` — resolver returned a non-`HandleNotFound` error (network, Bluesky outage). Surface a retry CTA, don't pretend it's pending.

These are the states real users will hit during setup.

sdd/bluesky-handle-setup/spec.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Data
+
+No new options required. We read `did` from the existing `atmosphere_connection` option that `Bluesky_Provider` already uses. The stored `handle` in that option is captured at OAuth time and goes stale once the user changes their handle on bsky.app, so it cannot be used to drive the "active" state.
+
+A single new option `fosse_bluesky_handle_setup_started` (boolean) tracks whether the user has clicked "Set up domain handle" so the CTA disappears once they're in the flow.
```
kraftbj: `planned-decisions.md` line 18 says this is an autoload-false option. Restate that here so the implementer of Task 4 doesn't accidentally autoload a value only read on one admin page.

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
+# Implementation Plan: Bluesky Handle Setup
+
+Based on: sdd/bluesky-handle-setup/spec.md
+
+## Progress
+
+- [ ] Task 1: Add `/.well-known/atproto-did` route handler
+- [ ] Task 2: Add verification check method
+- [ ] Task 2.5: Add `fetch_current_handle($did)` helper
```
kraftbj, line 9: All three of Task 1's commits already shipped on this branch (b49d940, 5d53ade, 7f3466d).

```suggestion
- [x] Task 1: Add `/.well-known/atproto-did` route handler
```

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+- [ ] Task 5: Add DNS TXT fallback display
+- [ ] Task 6: Unit tests
+- [ ] Task 6.5: Add Playwright E2E test for well-known route
+- [ ] Task 7: Update SDD documentation
+
+## Tasks
+
+### Task 1: Add `/.well-known/atproto-did` route handler
+- **Status**: In progress
+- **Files**: `src/Admin/class-bluesky-provider.php`
```
kraftbj, line 21:
```suggestion
- **Status**: ✅ Done (b49d940, 5d53ade, 7f3466d)
```

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+- **Do**:
+  1. Add `serve_atproto_did_well_known()` method. Hook to `init` priority 1 inside `register_hooks()`.
+  2. Match `$_SERVER['REQUEST_URI']` against `/^\/\.well-known\/atproto-did(\?|$)/`. If no match, return early.
```
kraftbj: Shipped code at `class-bluesky-provider.php:257-262` parses the URI properly rather than regex-matching:

```suggestion
  2. Parse `$_SERVER['REQUEST_URI']` with `wp_parse_url(..., PHP_URL_PATH)` and compare with strict equality against `/.well-known/atproto-did`. If no match, return early.
```

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+- **Depends on**: none
+
+### Task 2: Add verification check method
+- **Status**: Not started
+- **Files**: `src/Admin/class-bluesky-provider.php`
+- **Do**:
+  1. Add `check_handle_verification(string $domain): array` returning `['status', 'resolved_did', 'error']`.
+  2. Build `wp_remote_get` to `https://public.api.bsky.app/xrpc/com.atproto.identity.resolveHandle?handle=<domain>`.
+  3. Parse JSON. If `did` matches stored DID, status `verified`. If response has `did` but mismatched, status `mismatch`. If the resolver returns `HandleNotFound` (the common "domain not configured yet" case), status `not_found`. Other errors → `error`.
+  4. Cache result in transient `fosse_handle_check_<md5(domain . '|' . expected_did)>` for 5 minutes. Include the expected DID in the key so reconnecting a different account within the cache window doesn't reuse a stale `verified` result for the new DID.
```
kraftbj, line 41: This separator-form (`md5( $domain . '|' . $expected_did )`) is the correct one — see comment on Task 4 line 77 where the same key is written as `md5(domain.did)` without the separator. They have to match, and this version is collision-safe.

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+  - Unit test happy/missing/error branches via `pre_http_request` intercept.
+- **Depends on**: none
+
+### Task 3: Add domain-handle UI to Bluesky_Provider setup section
+- **Status**: Not started
+- **Files**: `src/Admin/class-bluesky-provider.php`
+- **Do**:
+  1. Extract a `render_domain_handle_subsection()` method called from the existing `render_setup_section()` when connected.
+  2. Implement five UI states from spec.md (hidden / CTA / pending / verified / active).
+  3. Use `wp_parse_url( home_url(), PHP_URL_HOST )` to extract the host as the suggested handle. This avoids producing invalid handles like `example.com/blog` on subdirectory installs.
```
kraftbj: Two gaps before this step is safe to ship:

1. **Eligibility gate.** Per the spec inline comments, only run domain-handle UI when `'' === trim( wp_parse_url( home_url(), PHP_URL_PATH ), '/' )`. On ineligible installs, render the explainer state instead — never reach this `wp_parse_url(... PHP_URL_HOST)` extraction.
2. **Lowercase the host.** ATProto handles must be lowercase. If the home URL ever stores capitalization (`Example.COM`), the resolver call silently fails. `strtolower()` the extracted host before passing it to `check_handle_verification()`. Worth a passing mention of IDN (`xn--…`) and port (`example.com:8080`) handling too — both produce non-handle output from the host extractor.

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+  - `composer run-script lint-php` passes.
+  - Playground: each UI state renders for a manually-set option combination.
+- **Depends on**: Task 1, Task 2, Task 2.5
+
+### Task 4: Add admin-post handlers
+- **Status**: Not started
+- **Files**: `src/Admin/class-bluesky-provider.php`
+- **Do**:
+  1. Add `admin_post_fosse_bluesky_start_handle_setup` handler. Verify nonce + capability. Set `fosse_bluesky_handle_setup_started` to `1`. Redirect back to FOSSE Setup with `settings-updated=true`.
+  2. Add `admin_post_fosse_bluesky_check_handle` handler. Verify nonce + capability. Bust both the verification transient (`fosse_handle_check_<md5(domain.did)>`) AND the current-handle transient (`fosse_handle_current_<md5(did)>`) so the UI reflects the post-handoff state immediately. Redirect back to FOSSE Setup with `settings-updated=true`.
```
kraftbj: Cache key here (`md5(domain.did)`) doesn't match Task 2 line 41 (`md5(domain . '|' . expected_did)`). They have to be identical or the bust here won't actually clear what Task 2 wrote. Use the `|`-separated form everywhere.

```suggestion
  2. Add `admin_post_fosse_bluesky_check_handle` handler. Verify nonce + capability. Bust both the verification transient (`fosse_handle_check_<md5(domain . '|' . expected_did)>`) AND the current-handle transient (`fosse_handle_current_<md5(did)>`) so the UI reflects the post-handoff state immediately. Redirect back to FOSSE Setup with `settings-updated=true`.
```

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+  - Playground: details element expands to show TXT record content. Copy button works.
+- **Depends on**: Task 3
+
+### Task 6: Unit tests
+- **Status**: Not started
+- **Files**: `tests/php/Admin/Bluesky_Handle_SetupTest.php`
+- **Do**:
+  1. Test `serve_atproto_did_well_known`: matches path, ignores other paths, respects filter, returns 404 when not connected.
+  2. Test `check_handle_verification`: each status branch via `pre_http_request` intercept. Caching honored.
+  3. Test admin-post handlers: nonce required, capability required, option toggled, transient busted, redirect target correct.
```
kraftbj: Missing test coverage for `fetch_current_handle` — Task 2.5 added a separate method (with happy / missing / error branches via `pre_http_request` intercept) that should be exercised here too. Looks like an oversight from when 2.5 was inserted. Also: tests should cover the eligibility gate (eligible vs ineligible based on a filtered `home_url`).

sdd/bluesky-handle-setup/plan.md
```diff
@@ -0,0 +1,127 @@
...
+  4. Test 3: When the `fosse_serve_atproto_did_well_known` filter returns `false`, the request returns 404 (verifies the disable filter works).
+- **Verify**:
+  - `pnpm exec playwright test tests/e2e/bluesky-handle.spec.ts` all green locally.
+- **Depends on**: Task 1
+
+### Task 7: Update SDD documentation
+- **Status**: Not started
+- **Files**: `sdd/bluesky-handle-setup/requirements.md`, `spec.md`, `plan.md`, `planned-decisions.md`
+- **Do**:
+  1. Update all four SDD documents to reflect as-built implementation. Add Done markers with PR ref.
```
kraftbj, line 127: AGENTS.md SDD workflow expects deviations to land in `implementation-notes.md`, but this SDD ships `planned-decisions.md` instead — a different document type that doesn't correspond to a step in the project workflow.

Two options:
1. Rename `planned-decisions.md` → `implementation-notes.md` once code lands, and have this task fold any pre-implementation decisions into it.
2. Keep `planned-decisions.md` as a separate pre-code artifact, and have this task explicitly create `implementation-notes.md` (a fifth file) once Tasks 1–6.5 are done.

Either is fine, but the plan should be explicit about which.

kraftbj (changes requested):
Changes requested. The architecture is sound — the verify-via-`resolveHandle` decision and the Verified-vs-Active distinction via `getProfile` are both well-considered. Most of what's below is doc hygiene plus a few state-machine gaps to close before Task 3 (UI) starts.

## Must-fix

1. **Plan status is stale.** Task 1's three commits (b49d940, 5d53ade, 7f3466d) already shipped on this branch, but `plan.md` still marks it "In progress" and the Progress checklist is unticked. Per AGENTS.md SDD convention, mark it ✅ Done with the commit refs.
2. **Cache-key spec disagrees with itself.** Task 2 uses `md5(domain . '|' . expected_did)`; Task 4 uses `md5(domain.did)`. Pick the `|`-delimited form — without a separator, `md5("foo.combined" . "did")` collides with `md5("foo" . ".combineddid")`.
3. **Eligibility / multisite gap.** The current spec assumes the WP install lives at the domain root. Subdirectory installs and subdirectory-multisite subsites can't use their URL as a handle (a path isn't part of an ATProto handle). The inline comments on `requirements.md` and `spec.md` propose a single check that handles every case:

   ```php
   '' === trim( wp_parse_url( home_url(), PHP_URL_PATH ), '/' )
   ```

   That naturally allows subdomain-multisite subsites and subdirectory-multisite *main* sites (both live at host-root) and excludes the path-bound cases. The UI needs an ineligibility state with an explainer instead of producing a broken handle.
4. **`planned-decisions.md` diverges from project SDD workflow.** AGENTS.md calls for `implementation-notes.md` to capture deviations once code lands. Either rename, or have Task 7 explicitly create `implementation-notes.md` alongside.

## High

5. Spec doesn't justify why FOSSE re-serves a route bundled Atmosphere already serves. The actual reason (FOSSE owns the opt-out filter; bundled-plugin policy forbids modifying Atmosphere) should be stated, plus a sentence on whether upstreaming the filter to Atmosphere was considered first.
6. Plan Task 1 describes a regex match but the shipped code uses `wp_parse_url(... PHP_URL_PATH)` + strict equality. Update the plan to match what shipped.
7. Plan Task 6 doesn't list tests for `fetch_current_handle` (added in Task 2.5).

## Medium

8. UI table only covers `verified`; `mismatch`, `not_found`, and `error` resolver statuses have no defined UI behavior. These are exactly the states users will hit during real setup.
9. Domain normalization (lowercase, IDN, port) not addressed before the resolver call.
10. Plan doesn't specify autoload-false for the new option (decisions doc says it should be).
11. "Single-site-owner flow" mentioned in requirements but not reaffirmed in spec.

## Strengths worth preserving

- The verify-via-Bluesky-resolver decision (Decision 1 in `planned-decisions.md`) avoids a real footgun.
- Distinguishing Verified from Active via `getProfile` instead of the stale stored handle is exactly right.
- The 5-minute transient with DID-keyed cache catches the account-reconnection edge case before it ships.

sdd/bluesky-handle-setup/requirements.md
RCowles: Addressed in f27a9c3. requirements.md now adds root-domain eligibility as requirement 6 and notes that the suggested handle comes from home_url host only when the URL path is empty.

sdd/bluesky-handle-setup/spec.md
RCowles: Addressed in f27a9c3. spec.md now has a Domain eligibility and normalization section before the UI state table, using the host-root check and documenting eligible vs path-bound install behavior.

sdd/bluesky-handle-setup/spec.md
RCowles: Addressed in f27a9c3. The UI state table now includes the path-bound/ineligible state before the connected setup states, with no setup buttons shown for that case.

sdd/bluesky-handle-setup/spec.md
RCowles: Addressed in f27a9c3. spec.md now explains why FOSSE re-serves the Atmosphere route: FOSSE owns the setup UX and opt-out contract, bundled-plugin policy keeps FOSSE-specific behavior out of bundled/, and a future generic Atmosphere opt-out hook could replace the local suppression shim. plan.md also now documents wp_parse_url(..., PHP_URL_PATH) plus strict equality instead of regex matching.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. The Progress checklist now marks Task 1 complete. The commit refs are the rebased local refs for this branch.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. Task 1 status is now done with the rebased commit refs for the branch: 396985f, 138404f, and bc43461. The later 8d6d01d commit tightened the connected-state behavior and tests.

sdd/bluesky-handle-setup/spec.md
RCowles: Addressed in f27a9c3. spec.md now states that fosse_bluesky_handle_setup_started is a boolean option with autoload disabled, and plan.md Task 4 specifies add_option(..., false) on first write.

sdd/bluesky-handle-setup/spec.md
RCowles: Addressed in f27a9c3. The UI state table now defines separate not_found, mismatch, and error states. mismatch shows both expected and resolved DID context; error keeps retry visible without presenting it as pending.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. plan.md Task 1 now describes parsing REQUEST_URI with wp_parse_url(..., PHP_URL_PATH) and strict equality against /.well-known/atproto-did.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. plan.md Task 4 now busts the verification transient with the same delimiter key shape as Task 2: fosse_handle_check_<md5(domain . "|" . did)>.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. plan.md Task 3 now gates setup controls on the host-root eligibility check, derives the candidate handle from home_url host, lowercases it, strips trailing dots, rejects explicit ports, and calls out IDN-to-ASCII conversion with error handling.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. The delimiter form is now the only cache-key contract in the SDD: md5( $domain . "|" . $expected_did ).

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. planned-decisions.md was replaced by implementation-notes.md, and Task 7 now lists implementation-notes.md as the SDD update target.

sdd/bluesky-handle-setup/plan.md
RCowles: Addressed in f27a9c3. plan.md Task 6 now includes fetch_current_handle coverage for happy, malformed, HTTP, WP_Error, and caching behavior, plus domain normalization and eligibility coverage including path-bound installs and IDN support.

RCowles:
Review summary follow-up: I replied in-thread on the inline items. The summary-only single-site-owner note is also addressed in f27a9c3: spec.md now reaffirms this as a single-site-owner setup flow, and requirements.md keeps per-user/per-author handles out of scope.

kraftbj:
Pushed 0f8a29e addressing the critical and high-priority items from the latest review pass:

**Code (`src/Admin/class-bluesky-provider.php`)**
- DID body now goes through `esc_html()` to match bundled Atmosphere's upstream pattern and guard against malformed stored values.
- `$_SERVER['REQUEST_URI']` read raw (no `sanitize_text_field`) so encoded sequences don't get normalized into surprising matches on the strict path-equality check.
- "Atmosphere not loaded" now returns `null` (decline to handle, falls through to normal WordPress 404) instead of returning a 404 — a structural integrity failure no longer looks like a user-facing not-configured state.
- Dropped the redundant `function_exists( '\Atmosphere\get_connection' )` check (defined in the same file as `is_connected`).
- PHPDoc on `serve_atproto_did_well_known` rewritten to describe all three return paths; PHPDoc on `maybe_suppress_atmosphere_well_known` now mentions the 404 flagging. Return-shape annotation tightened to `status:200|404`.

**Tests (`tests/php/Admin/Bluesky_ProviderTest.php`)**
- Pinned the new hooks at priority `1` in `test_register_hooks_adds_actions` (was `assertNotFalse`, which accepts any priority — a typo dropping the explicit `1` would have made the suppression race intermittent).
- Added three tests for `maybe_suppress_atmosphere_well_known`: query-var miss, filter-true (FOSSE serves it), and opted-out (clears query var + 404s).

**Deferred to Task 6**

A few test-quality items from the review pass don't change behavior and are better folded into the broader Task 6 test sweep than into a Task 1 follow-up:

- HTTP side-effect coverage on the public `serve_atproto_did_well_known` (body content, no-trailing-newline, `Content-Type`) — needs a small testability refactor (wrap `exit` in a `terminate()` shim or use `runInSeparateProcess`).
- Path-match edge-case tests (trailing slash, case sensitivity, percent-encoded sequences).
- Renaming `test_atproto_did_well_known_response_requires_connected_atmosphere` to better describe what it actually exercises (the `is_connected()` branch, not the empty-DID branch).

Lint clean, all 117 PHP tests passing.

kraftbj:
Pushed d56d7aa addressing two findings from the codex re-review on 0f8a29e:

- The `esc_html()` I added to the DID body in 0f8a29e contradicted the spec ("plain-text", "No HTML") and would have corrupted any DID containing `&`. Replaced with raw output plus an upstream check: the helper now validates the stored DID against AT Proto syntax (`did:<method>:<id>` with id restricted to ASCII alphanumerics + `._:%-`) and returns 404 if the value doesn't match. Defense in depth at the right layer — at-write-time validation rather than at-render-time escaping in the wrong content type. Atmosphere's upstream pattern is also wrong on principle (text/plain shouldn't go through esc_html) but harmless since valid DIDs never contain HTML chars.
- The three suppression-hook tests now capture `status_header` via the filter and assert on it: the no-op tests assert it's never called, and the opt-out test asserts `status_header( 404 )` was actually sent. Without this, a regression dropping `status_header(404)` would still have passed all three tests.

New test added: `test_atproto_did_well_known_response_rejects_malformed_did` exercises the syntax check with a stored DID containing a newline + HTML payload.

Lint clean, all 118 PHP tests passing.

kraftbj:
Pushed 5f4a25c addressing the regex-anchor issue both reviewers flagged on d56d7aa:

PHP's `$` anchor allows a single trailing `\n`, so a stored value like `did:plc:test123\n` passed validation and would have emitted a stray newline byte — violating the spec's "no trailing newline" requirement. Switched to `\A` + `\z` for absolute string boundaries, and tightened the tail to require a non-`:`/non-`%` final character. New regression test (`test_atproto_did_well_known_response_rejects_did_with_trailing_newline`) pins the trailing-newline rejection.

Lint clean, all 119 PHP tests passing.

kraftbj (dismissed):
LGTM — Task 1 implementation is in good shape after this round of fixes.

The DID render layer now has tight defense in depth (AT Proto syntax check at the helper, raw `text/plain` body, no HTML-escape contradiction), the suppression hook has real behavioral coverage including `status_header(404)` capture, and the hooks are pinned to priority 1 so a refactor losing the explicit priority would fail loudly instead of racing with bundled Atmosphere.

Codex flagged one residual nicety on 5f4a25c — the regex accepts a bare `%` not followed by two hex digits, which is not strict per the DID percent-encoding spec but is fine as a defense-in-depth byte filter (the goal is rejecting newlines/control chars/HTML, not full syntax validation). Worth tightening if a future DID parser is added; not blocking.

Other test-quality items from earlier review rounds (HTTP side-effect coverage on the public method, path-match edge cases, a couple of test renames) are documented in the PR comment thread and ride along with the broader Task 6 test pass.

PR #47, opened 2026-04-30
Add Bluesky connect flow to onboarding wizard

RCowles:
## Summary
- Replace the Bluesky onboarding placeholder with provider-backed live connect, connected, and unavailable states.
- Preserve wizard-origin OAuth return context so callbacks return to the Bluesky wizard step.
- Keep this PR scoped to the current wizard order; reordering the wizard can be handled in a follow-up after this lands.

## Testing
- `composer run-script lint-php`
- `composer run-script test-php`
- `pnpm run format:check`
- `pnpm run lint`
- `pnpm exec playwright test tests/e2e/onboarding-wizard.spec.ts`

### DIFF
--- a/sdd/onboarding-setup-ux/plan.md
+++ b/sdd/onboarding-setup-ux/plan.md
@@ -10,6 +10,7 @@ Based on: sdd/onboarding-setup-ux/spec.md
 - [x] Task 4: Open upstream Atmosphere PRs
 - [x] Task 5: Create Bluesky_Provider
 - [x] Task 5.5: Create first-run onboarding wizard
+- [ ] Task 5.6: Replace Bluesky wizard placeholder with live connect form
 - [x] Task 6: Create Status_Page with provider status cards
 - [x] Task 7: Add admin CSS
 - [x] Task 8: Write tests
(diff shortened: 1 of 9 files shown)


kraftbj:
I'm going to push up a few changes here in a minute.

kraftbj:
Pushed four follow-up commits addressing the review feedback. Grouped by risk so each one is easy to revert independently.

### `b364d7c` — fix: harden Bluesky OAuth callback (state binding, persist check)

The big one. The wizard return-context transient was being consumed unconditionally after `Atmosphere\OAuth\Client::handle_callback()`, so a stale or replayed `?code&state=bad` could strip the wizard marker before the legitimate callback arrived — leaving onboarding stuck on the Setup page after the user finished Bluesky auth. The marker is now keyed `[context, state]`, and `consume_oauth_return_context()` only deletes when the inbound `state` matches via `hash_equals`. A non-matching callback returns `''` and leaves the marker for the real callback.

While in there, two adjacent silent-failure shapes:

- After `handle_callback()` returns true, re-check `\Atmosphere\is_connected()`. A lost option write (DB error, hostile filter) used to surface as a green success notice followed by a wizard re-rendering the connect form.
- A missing `Atmosphere\Publisher::sync_publication` no longer evaluates as `null is_wp_error → success`. It now emits a `warning` notice telling the user publication setup is unavailable in the current Atmosphere version.

### `e1813d0` — fix: verify Bluesky disconnect, cover OAuth callback success branches

`handle_disconnect` was claiming "Disconnected" regardless of whether the option actually went away (`Client::disconnect()` returns void, `delete_option()` failures were invisible) and was leaking any in-flight wizard return marker. Now it calls `forget_oauth_return_context()` and re-reads `is_connected()` after the disconnect, surfacing an error notice if the option survived.

The PR-test agent flagged that the success / `is_connected`-false / `sync_publication`-WP_Error branches of `handle_oauth_callback` were untested — only the WP_Error path of `handle_callback` was exercised. Added a fixture (`fake_atmosphere_oauth_environment` + `intercept_token_endpoint_success` + `intercept_pds_create_record_*`) that seeds Atmosphere's transients and short-circuits `wp_remote_post`, so we can assert on the actual notice type emitted in each branch.

### `d29527d` — fix: tighten Bluesky availability, validation, and callback guards

Five smaller items rolled up:

- The wizard's `new Bluesky_Provider()` registry fallback is gone. If something prevents registration, the connect form would have posted to an `admin-post.php` action with no handler attached. Now the wizard renders the unavailable notice — a worse-looking UI but at least a debuggable signal instead of a silent 404.
- `is_available()` now also requires `\Atmosphere\get_connection` and `\Atmosphere\is_connected`. `get_status()` was guarding those with `function_exists` defaults that disagreed with `is_available()`; with the tighter check, the inline guards are gone.
- An OAuth callback with exactly one of `code` / `state` present is now an error notice instead of a silent return. Both missing is still a no-op (ordinary admin page hit).
- Handle input gets a domain-shape regex check before reaching `Client::authorize()`, so `@alice` or `alice bsky.social` returns "That doesn't look like a Bluesky handle" instead of whatever raw error the upstream resolver coughs up.
- New test extracts the wizard form's `_wpnonce` and validates it against `fosse_connect_bluesky` — a typo'd action would now fail the test instead of 403'ing in production.

### `dbc138c` — refactor: tighten Bluesky return-context plumbing and test catches

Type-design / code-cleanup pass. `RETURN_CONTEXT_FIELD` and `RETURN_CONTEXT_WIZARD` are public so the wizard form references them directly instead of duplicating the literal strings. A single `normalize_return_context()` chokepoint replaces the five `=== self::RETURN_CONTEXT_WIZARD ? ... : ''` ternaries. `get_oauth_return_transient_key()` returns `''` when `get_current_user_id() === 0` so the per-user namespace can't collapse into a shared anonymous key. Tests now `catch RedirectFired` instead of `\Exception` so unrelated runtime exceptions surface instead of being swallowed.

---

Verification: 135 / 135 PHPUnit, PHPCS clean. The deeper success-path tests added in commit two now exercise the `success` / `warning` / `error` notice types end-to-end via Atmosphere, which closes the "manual verification only" gap the SDD plan called out.

PR #48, opened 2026-04-30
Refresh bundled activitypub and atmosphere from upstream trunk

kraftbj:
## Summary

Re-vendors `bundled/activitypub/` and `bundled/atmosphere/` from upstream trunks via `tools/sync-bundled.sh`. Both backends shipped as production builds.

- **ActivityPub**: 8.1.1 → 8.2.0 (release plus a few post-release dependabot bumps). Pulls in security hardening across OAuth, HTTP signatures, SSRF/DPoP, and inbox handling. Production `build/` assets are committed in upstream and refreshed accordingly.
- **Atmosphere**: ~10 commits forward (long-form `atmosphere_long_form_composition` filter that the existing `Long_Form_Strategy` projector already drives, new `class-post-types.php`, plugin bootstrap/teardown cleanup, native-OpenSSL DPoP, OAuth `redirect_uri` filter). Production `vendor/` rebuilt by the sync script with `composer install --no-dev --optimize-autoloader` before rsync.

## Verification

- `composer run-script lint-php` — clean (PHPCS, Jetpack ruleset)
- `pnpm run lint` — clean (ESLint)
- `pnpm run format:check` — clean (Prettier)
- `composer run-script test-php` — 110 tests / 227 assertions, all green

## Notes / follow-ups (not blocking)

- Atmosphere's upstream `composer.lock` is stale relative to `composer.json` (composer warned during the sync). Since the JWT library swap to native OpenSSL, atmosphere's custom autoloader no longer references `vendor/autoload.php` anywhere — but the lock still pins `web-token/jwt-library`, `spomky-labs/pki-framework`, `brick/math`, and `psr/clock`, so we're shipping ~1.8 MB of unused JWT code in `bundled/atmosphere/vendor/`. Per the upstream-contribution policy in `AGENTS.md`, the fix lives upstream (run `composer update --lock` on atmosphere trunk); we'd then re-sync here.
- `tools/bundled-excludes.txt`'s comment claiming "Atmosphere's vendor/autoload.php is required at plugin load time" is now out of date for the same reason.


kraftbj:
Correction on the upstream-lockfile note from the PR description: the atmosphere upstream agent debunked it, and they were right.

`wordpress-atmosphere`'s `composer.lock` is **gitignored** upstream — there's no tracked lock to be stale. The dead JWT packages in the original bundled snapshot here came from a stale, untracked `composer.lock` in *my local atmosphere checkout* (left over from a pre-#23 `composer install` when `web-token/jwt-library` was a runtime dep). When `tools/sync-bundled.sh` ran `composer install --no-dev` against that stale lock, it kept those packages and rsync'd them into `bundled/atmosphere/vendor/`.

Pushed `5faf20a` to fix it locally without involving upstream:

- Scrubbed `~/code/wordpress-atmosphere/{composer.lock,vendor}` and re-ran `composer install --no-dev`. Atmosphere has zero runtime composer deps now (custom autoloader since #23), so the install is a no-op and `vendor/` contains only composer's own autoload scaffolding.
- Re-ran `tools/sync-bundled.sh` — `bundled/atmosphere/vendor/` is now `autoload.php` + `composer/` only (~50 KB instead of ~1.8 MB). 544 files / ~50k lines removed in this commit.
- Updated the comment in `tools/bundled-excludes.txt` that incorrectly claimed atmosphere's `vendor/autoload.php` was required at plugin load time.

Lint and the 110-test PHPUnit suite still pass.

PR #49, opened 2026-04-30
Clarify ActivityPub profile mode settings

### LINKED ISSUE #37: Settings: explain "Author profile" vs "Blog profile" and how to set blog profile
## Context

Surfaced while reviewing the Settings page after upgrading an existing FOSSE install (so the wizard didn't run). Observed on `trunk`.

## Problem

The actor-mode control (Author profile vs Blog profile) has no inline explanation:

- A user landing on the Settings page can't tell what each mode actually does — author-per-user vs single site-wide actor, addressability differences, who appears as the publisher in remote networks, etc.
- From the Settings screen there's **no obvious path to set the Blog profile** if that's the mode they want.
- It's unclear what happens to existing followers/posts/data if someone **changes modes after activation**. (I'm not sure the bundled AP plugin documents this either, so it may need to land upstream too.)

## Suggested direction

- Inline help text under each mode option describing what it does in plain terms.
- Surface (or link to) the blog-profile configuration directly from this screen instead of requiring users to know where it lives.
- Document the consequence of switching modes after the fact (do followers carry? do post URIs change? does addressing break?). One short paragraph plus a help-center link.
- Coordinate any upstream gaps with the bundled ActivityPub plugin so the explanation lives somewhere durable.

## Why it matters

This is the foundational identity choice a FOSSE site makes. If it's unclear at setup, users either pick blindly or stall. Clear help text on the settings surface bridges the gap until/unless the wizard handles it for new installs.

RCowles:
## Summary
- Adds inline help text for ActivityPub actor/profile modes on the FOSSE Settings page.
- Links directly to ActivityPub Blog Profile settings.
- Adds a fieldset-wide note that changing modes does not move followers between profiles.
- Adds render coverage for the new help text and accessible descriptions.

## Testing
- pnpm test
- pnpm run format:check
- pnpm run lint
- git diff --check

Fixes #37.

### DIFF
--- a/src/Admin/class-ap-provider.php
+++ b/src/Admin/class-ap-provider.php
@@ -116,13 +116,14 @@ class AP_Provider implements Connection_Provider {
 						<th scope="row"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></th>
 						<td>
 							<fieldset aria-describedby="fosse-activitypub-actor-mode-note">
+								<legend class="screen-reader-text"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></legend>
 								<label>
 									<input
 										type="radio"
 										id="fosse-activitypub-actor-mode-actor"
 										name="activitypub_actor_mode"
 										value="actor"
-										aria-describedby="fosse-activitypub-actor-mode-actor-desc"
+										aria-describedby="fosse-activitypub-actor-mode-actor-desc fosse-activitypub-actor-mode-note"
 										<?php checked( 'actor', $actor_mode ); ?>
 									/>
 									<?php esc_html_e( 'Author profiles', 'fosse' ); ?>
@@ -136,7 +137,7 @@ class AP_Provider implements Connection_Provider {
 										id="fosse-activitypub-actor-mode-blog"
 										name="activitypub_actor_mode"
 										value="blog"
-										aria-describedby="fosse-activitypub-actor-mode-blog-desc"
+										aria-describedby="fosse-activitypub-actor-mode-blog-desc fosse-activitypub-actor-mode-note"
 										<?php checked( 'blog', $actor_mode ); ?>
 									/>
 									<?php esc_html_e( 'Blog profile', 'fosse' ); ?>
@@ -150,7 +151,7 @@ class AP_Provider implements Connection_Provider {
 										id="fosse-activitypub-actor-mode-actor-blog"
 										name="activitypub_actor_mode"
 										value="actor_blog"
-										aria-describedby="fosse-activitypub-actor-mode-actor-blog-desc"
+										aria-describedby="fosse-activitypub-actor-mode-actor-blog-desc fosse-activitypub-actor-mode-note"
 										<?php checked( 'actor_blog', $actor_mode ); ?>
 									/>
 									<?php esc_html_e( 'Both', 'fosse' ); ?>
@@ -158,8 +159,8 @@ class AP_Provider implements Connection_Provider {
 								<p id="fosse-activitypub-actor-mode-actor-blog-desc" class="description">
 									<?php esc_html_e( 'Authors keep individual profiles, and the site also has its own blog profile. People can follow either.', 'fosse' ); ?>
 								</p>
-								<div id="fosse-activitypub-actor-mode-note" class="fosse-activitypub-actor-mode-note">
-									<p class="description">
+								<div class="fosse-activitypub-actor-mode-note">
+									<p id="fosse-activitypub-actor-mode-note" class="description">
 										<?php esc_html_e( 'Changing modes does not move followers between profiles. Future posts publish from the profiles enabled by the selected mode.', 'fosse' ); ?>
 									</p>
 									<p class="description">
--- a/src/Admin/class-ap-provider.php
+++ b/src/Admin/class-ap-provider.php
@@ -115,7 +115,7 @@ class AP_Provider implements Connection_Provider {
 					<tr>
 						<th scope="row"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></th>
 						<td>
-							<fieldset aria-describedby="fosse-activitypub-actor-mode-note">
+							<fieldset>
 								<legend class="screen-reader-text"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></legend>
 								<label>
 									<input
@@ -167,7 +167,7 @@ class AP_Provider implements Connection_Provider {
 										<?php
 										echo wp_kses(
 											sprintf(
-												/* translators: %s: link to ActivityPub blog profile settings. */
+												/* translators: %s: anchor link reading "Blog profile settings" pointing to the ActivityPub blog profile tab. */
 												__( 'Configure the site-wide blog profile name, image, and description in %s.', 'fosse' ),
 												'<a href="' . esc_url( admin_url( 'options-general.php?page=activitypub&tab=blog-profile' ) ) . '">' . esc_html__( 'Blog profile settings', 'fosse' ) . '</a>'
 											),
--- a/src/Admin/class-ap-provider.php
+++ b/src/Admin/class-ap-provider.php
@@ -115,19 +115,70 @@ class AP_Provider implements Connection_Provider {
 					<tr>
 						<th scope="row"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></th>
 						<td>
-							<fieldset>
+							<fieldset aria-describedby="fosse-activitypub-actor-mode-note">
 								<label>
-									<input type="radio" name="activitypub_actor_mode" value="actor" <?php checked( 'actor', $actor_mode ); ?> />
+									<input
+										type="radio"
+										id="fosse-activitypub-actor-mode-actor"
+										name="activitypub_actor_mode"
+										value="actor"
+										aria-describedby="fosse-activitypub-actor-mode-actor-desc"
+										<?php checked( 'actor', $actor_mode ); ?>
+									/>
 									<?php esc_html_e( 'Author profiles', 'fosse' ); ?>
-								</label><br />
+								</label>
+								<p id="fosse-activitypub-actor-mode-actor-desc" class="description">
+									<?php esc_html_e( 'Each WordPress author publishes from their own fediverse profile. People follow individual authors, and posts appear under each author\'s name.', 'fosse' ); ?>
+								</p>
 								<label>
-									<input type="radio" name="activitypub_actor_mode" value="blog" <?php checked( 'blog', $actor_mode ); ?> />
+									<input
+										type="radio"
+										id="fosse-activitypub-actor-mode-blog"
+										name="activitypub_actor_mode"
+										value="blog"
+										aria-describedby="fosse-activitypub-actor-mode-blog-desc"
+										<?php checked( 'blog', $actor_mode ); ?>
+									/>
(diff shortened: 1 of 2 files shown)


tests/php/Admin/AP_ProviderTest.php
```diff
@@ -133,6 +133,42 @@ public function test_render_status_card_has_manage_settings_link() {
...
+		);
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-actor-desc"', $output );
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-blog-desc"', $output );
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-actor-blog-desc"', $output );
+		$this->assertStringContainsString( '<div id="fosse-activitypub-actor-mode-note" class="fosse-activitypub-actor-mode-note">', $output );
+		$this->assertMatchesRegularExpression(
+			'~<a href="[^"]*options-general\.php\?page=activitypub(?:&#038;|&amp;|&)tab=blog-profile">Blog profile settings</a>~',
+			$output
+		);
+	}
```
kraftbj, line 168: **Coverage gap on `checked()` wiring.** The PR reformatted each radio's `<?php checked(...); ?>` into multi-line attributes, but no test in this file ever calls `render_setup_section()` with a non-default `activitypub_actor_mode` and verifies the matching radio is selected. A typo or stray whitespace in the new layout would silently always render "Author profiles" selected, and nothing would catch it.

Worth a small parametric test over the three modes (`actor`, `blog`, `actor_blog`) that sets the option and asserts the right input emits `checked='checked'` while the others don't. This is the most important comment in this review.

tests/php/Admin/AP_ProviderTest.php
```diff
@@ -133,6 +133,42 @@ public function test_render_status_card_has_manage_settings_link() {
...
+		$this->assertMatchesRegularExpression(
+			'~<input[^>]+id="fosse-activitypub-actor-mode-actor-blog"[^>]+aria-describedby="fosse-activitypub-actor-mode-actor-blog-desc"[^>]+/>~',
+			$output
+		);
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-actor-desc"', $output );
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-blog-desc"', $output );
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-actor-blog-desc"', $output );
+		$this->assertStringContainsString( '<div id="fosse-activitypub-actor-mode-note" class="fosse-activitypub-actor-mode-note">', $output );
+		$this->assertMatchesRegularExpression(
+			'~<a href="[^"]*options-general\.php\?page=activitypub(?:&#038;|&amp;|&)tab=blog-profile">Blog profile settings</a>~',
```
kraftbj: Suggest tightening to `(?:&#038;|&amp;)` — drop the bare `&`. `esc_url()` always emits one of the entity forms, so accepting bare `&` here means a future regression that swaps `esc_url(...)` for `admin_url(...)` directly would slip past this assertion. No false-fail risk from tightening since real output will never contain a bare `&`.

tests/php/Admin/AP_ProviderTest.php
```diff
@@ -133,6 +133,42 @@ public function test_render_status_card_has_manage_settings_link() {
...
+			$output
+		);
+		$this->assertMatchesRegularExpression(
+			'~<input[^>]+id="fosse-activitypub-actor-mode-actor-blog"[^>]+aria-describedby="fosse-activitypub-actor-mode-actor-blog-desc"[^>]+/>~',
+			$output
+		);
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-actor-desc"', $output );
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-blog-desc"', $output );
+		$this->assertStringContainsString( 'id="fosse-activitypub-actor-mode-actor-blog-desc"', $output );
+		$this->assertStringContainsString( '<div id="fosse-activitypub-actor-mode-note" class="fosse-activitypub-actor-mode-note">', $output );
```
kraftbj: The three `id="..."` `assertStringContainsString` calls on lines 162–164 plus this exact-form `<div id="..." class="...">` assertion duplicate what the regex assertions above already imply about the a11y wiring, and pin the spelling and class of internal IDs that aren't part of any user-facing contract. Renaming the IDs (e.g. `fosse-activitypub-actor-mode-*` → `fosse-ap-actor-mode-*`) would break 7 assertions on a no-op refactor.

Safe to drop these four assertions — the three regexes above carry the real behavioral contract (input's `aria-describedby` value matches a real id in the output).

src/Admin/class-ap-provider.php
```diff
@@ -115,19 +115,70 @@ public function render_setup_section(): void {
...
+										value="actor_blog"
+										aria-describedby="fosse-activitypub-actor-mode-actor-blog-desc"
+										<?php checked( 'actor_blog', $actor_mode ); ?>
+									/>
 									<?php esc_html_e( 'Both', 'fosse' ); ?>
 								</label>
+								<p id="fosse-activitypub-actor-mode-actor-blog-desc" class="description">
+									<?php esc_html_e( 'Authors keep individual profiles, and the site also has its own blog profile. People can follow either.', 'fosse' ); ?>
+								</p>
+								<div id="fosse-activitypub-actor-mode-note" class="fosse-activitypub-actor-mode-note">
```
kraftbj: Minor a11y nit: this `<div>` is the target of the fieldset's `aria-describedby` (line 118), but it lives *inside* that same fieldset. Valid per ARIA, but some AT/screen-reader combinations double-announce — once when entering the fieldset, again when reaching the div via normal reading order.

Two cleaner options:
- Move this `<div>` to be a sibling *after* `</fieldset>`, keeping the `aria-describedby` reference.
- Drop the fieldset-level `aria-describedby` entirely; the `class="description"` paragraphs already convey the same content via reading order.

src/Admin/class-ap-provider.php
```diff
@@ -115,19 +115,70 @@ public function render_setup_section(): void {
 					<tr>
 						<th scope="row"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></th>
 						<td>
-							<fieldset>
+							<fieldset aria-describedby="fosse-activitypub-actor-mode-note">
```
kraftbj: Pre-existing, but flagging since this PR is enriching the fieldset for a11y anyway: the fieldset has no `<legend>`. The `<th scope="row">Actor Mode</th>` is doing visual-label duty but isn't programmatically associated with `<fieldset>`, so the radio group has no accessible name. A visually-hidden `<legend>Actor Mode</legend>` here would close that gap. Optional — not a regression.

src/Admin/class-ap-provider.php
```diff
@@ -115,19 +115,70 @@ public function render_setup_section(): void {
...
+								</p>
+								<div id="fosse-activitypub-actor-mode-note" class="fosse-activitypub-actor-mode-note">
+									<p class="description">
+										<?php esc_html_e( 'Changing modes does not move followers between profiles. Future posts publish from the profiles enabled by the selected mode.', 'fosse' ); ?>
+									</p>
+									<p class="description">
+										<?php
+										echo wp_kses(
+											sprintf(
+												/* translators: %s: link to ActivityPub blog profile settings. */
```
kraftbj: Polish: many languages need to bend grammar around the visible anchor text, so translators benefit from knowing what the link reads. Consider: `/* translators: %s: anchor link reading "Blog profile settings" pointing to the ActivityPub blog profile tab. */`. Strictly optional.

kraftbj (commented):
Review notes from local analysis — none are blockers, but the test-coverage gap on `checked()` wiring is worth closing before merge.

kraftbj (dismissed):
Approving — the remaining review comments are non-blocking suggestions (test coverage gap on `checked()` wiring, regex tightening, brittle ID-pinning, translator-comment polish). Worth picking up in a follow-up but not gating this PR.

src/Admin/class-ap-provider.php
```diff
@@ -115,19 +115,70 @@ public function render_setup_section(): void {
 					<tr>
 						<th scope="row"><?php esc_html_e( 'Actor Mode', 'fosse' ); ?></th>
 						<td>
-							<fieldset>
+							<fieldset aria-describedby="fosse-activitypub-actor-mode-note">
 								<label>
```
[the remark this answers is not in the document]
RCowles: Addressed in `980f711`: the fieldset now includes `<legend class="screen-reader-text">Actor Mode</legend>`. `test_render_setup_section_explains_actor_modes()` asserts the legend is rendered.

tests/php/Admin/AP_ProviderTest.php
RCowles: Addressed in `9092175`: the href assertion now accepts only `&#038;` or `&amp;`, not a raw ampersand.

tests/php/Admin/AP_ProviderTest.php
RCowles: Addressed in `9092175`: added `test_render_setup_section_checks_selected_actor_mode()` to iterate `actor`, `blog`, and `actor_blog`, set `activitypub_actor_mode`, and assert only the matching radio input contains `checked='checked'`.

src/Admin/class-ap-provider.php
```diff
@@ -115,19 +115,70 @@ public function render_setup_section(): void {
...
-									<input type="radio" name="activitypub_actor_mode" value="actor_blog" <?php checked( 'actor_blog', $actor_mode ); ?> />
+									<input
+										type="radio"
+										id="fosse-activitypub-actor-mode-actor-blog"
+										name="activitypub_actor_mode"
+										value="actor_blog"
+										aria-describedby="fosse-activitypub-actor-mode-actor-blog-desc"
+										<?php checked( 'actor_blog', $actor_mode ); ?>
+									/>
```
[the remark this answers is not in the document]
RCowles: Addressed in `980f711` and retained after `9092175`: each radio control now describes itself with its option-specific description plus `fosse-activitypub-actor-mode-note`. The note ID is on the follower-mode paragraph only, so the Blog Profile link is not read as part of every radio description. Covered by `test_render_setup_section_explains_actor_modes()`.

tests/php/Admin/AP_ProviderTest.php
RCowles: Addressed in `9092175`: removed the standalone description-ID assertions and the exact wrapper assertion. The test now keeps the behavioral checks for each radio control description and the shared note paragraph.

src/Admin/class-ap-provider.php
RCowles: Addressed in `9092175`: dropped the fieldset-level `aria-describedby`. Each radio control still includes `fosse-activitypub-actor-mode-note` in its own `aria-describedby`, so the shared note is available on focus without also describing the fieldset.

src/Admin/class-ap-provider.php
RCowles: Addressed in `980f711`: added `<legend class="screen-reader-text">Actor Mode</legend>` inside the fieldset, and the render test asserts it is present.

src/Admin/class-ap-provider.php
RCowles: Addressed in `9092175`: expanded the translator note to say the inserted anchor reads "Blog profile settings" and points to the ActivityPub blog profile tab.

PR #99, opened 2026-05-05
Wizard: warm up onboarding visuals

RCowles:
## Summary
- give the onboarding wizard a warmer shell with a soft background wash, wider content column, and more breathing room around the progress rail
- restyle the destination and identity cards with rounded panels, subtle elevation, hover states, and stronger selected/focus states
- add icons to the destination cards so the first step feels less sparse without changing wizard behavior

## Testing
- composer run-script test-php -- --filter Onboarding_WizardTest
- pnpm exec prettier --check src/Admin/assets/css/admin.css
- pnpm run lint
- git diff --check

## Notes
- This is stacked on #88 / `onboarding-wizard-ux-review`; retarget to `trunk` after #88 merges if needed.
- Full `pnpm run format:check` is noisy locally because `.air/mcp.json` is untracked on the stacked base branch; the touched CSS file passes Prettier directly.

### DIFF
--- a/sdd/destination-first-onboarding-wizard/plan.md
+++ b/sdd/destination-first-onboarding-wizard/plan.md
@@ -12,13 +12,13 @@
 
 ## Progress
 
-- [ ] Task 1: Add destination state and save routing
-- [ ] Task 2: Render the destination-card first step
-- [ ] Task 3: Update progress, identity copy, and content grouping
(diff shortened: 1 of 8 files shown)


RCowles:
Closing this draft so we can restart wizard UI polish from a clean trunk base.

PR #116, opened 2026-05-08
Polish onboarding wizard styling

RCowles:
Summary:
- Refines onboarding wizard styling and progress treatment.
- Adds a small client-side styling toggle scoped to the wizard.
- Covers the render and toggle behavior with focused tests.

Testing:
- pnpm run format:check
- pnpm run lint
- pnpm test

### DIFF
--- a/src/Admin/assets/js/wizard-lizard.js
+++ b/src/Admin/assets/js/wizard-lizard.js
@@ -1,0 +1,74 @@
+/**
+ * Wizard lizard easter egg.
+ *
+ * Clicking the deliberately quiet lizard button adds a CSS class that
+ * turns the setup flow into the Onboarding Lizard without changing any
+ * form behavior. Session storage keeps the theme alive across wizard steps.
+ */
+( function () {
+	'use strict';
+
+	const THEME_CLASS = 'is-lizard-themed';
+	const STORAGE_KEY = 'fosseWizardLizardTheme';
+	const TOGGLE_SELECTOR = '[data-fosse-lizard-toggle]';
+	const WIZARD_SELECTOR = '.fosse-wizard';
+
+	function readStoredTheme() {
+		try {
+			return window.sessionStorage.getItem( STORAGE_KEY ) === '1';
+		} catch {
+			return false;
+		}
+	}
+
+	function storeTheme( enabled ) {
+		try {
+			if ( enabled ) {
+				window.sessionStorage.setItem( STORAGE_KEY, '1' );
+			} else {
+				window.sessionStorage.removeItem( STORAGE_KEY );
+			}
+		} catch {
+			// Storage can be unavailable in restricted browser contexts.
+		}
+	}
+
+	function applyTheme( wizard, toggle, enabled ) {
+		wizard.classList.toggle( THEME_CLASS, enabled );
+		toggle.setAttribute( 'aria-pressed', enabled ? 'true' : 'false' );
+	}
+
+	function init( root ) {
+		const scope = root || document;
+		const wizard = scope.querySelector( WIZARD_SELECTOR );
+
+		if ( ! wizard ) {
+			return;
+		}
+
+		const toggle = wizard.querySelector( TOGGLE_SELECTOR );
+
+		if ( ! toggle ) {
+			return;
+		}
+
+		if ( readStoredTheme() ) {
+			applyTheme( wizard, toggle, true );
+		}
+
+		toggle.addEventListener( 'click', function () {
+			const enabled = ! wizard.classList.contains( THEME_CLASS );
+
+			applyTheme( wizard, toggle, enabled );
+			storeTheme( enabled );
+		} );
+	}
+
+	if ( document.readyState === 'loading' ) {
+		document.addEventListener( 'DOMContentLoaded', function () {
+			init();
+		} );
+	} else {
+		init();
+	}
+} )();
--- a/src/Admin/class-onboarding-wizard.php
+++ b/src/Admin/class-onboarding-wizard.php
@@ -155,6 +155,7 @@ class Onboarding_Wizard {
 		?>
 		<div class="wrap fosse-wizard">
 		<?php
+		self::render_lizard_toggle();
 
 		if ( ! self::is_activitypub_available() ) {
 			self::render_unavailable_notice();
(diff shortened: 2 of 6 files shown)


src/Admin/class-onboarding-wizard.php
```diff
@@ -155,6 +155,7 @@ public static function render(): void {
 		?>
 		<div class="wrap fosse-wizard">
 		<?php
+		self::render_lizard_toggle();
```
kraftbj, line 158: The toggle renders before the `is_activitypub_available()` early-return below, so the easter egg also appears on the AP-unavailable notice screen. Looks intentional (and visually harmless given the wrapper has no themed content on that path), but `test_render_shows_degraded_notice_when_activitypub_unavailable` doesn't assert `data-fosse-lizard-toggle`, so a refactor that moved this call inside the AP-available branch would silently regress with no test failure. A one-liner assertion in that test would pin the contract.

tests/js/wizard-lizard.test.js
```diff
@@ -0,0 +1,60 @@
...
+			</div>
+		`;
+	}
+
+	function loadScript() {
+		jest.isolateModules( () => {
+			require( '../../src/Admin/assets/js/wizard-lizard.js' );
+		} );
+
+		document.dispatchEvent( new Event( 'DOMContentLoaded' ) );
```
kraftbj, line 17: This `dispatchEvent` is a no-op as written — `jest-environment-jsdom` initializes `document.readyState` to `'complete'`, so the IIFE in `wizard-lizard.js` always takes the else-branch and calls `init()` synchronously at `require()` time. No `DOMContentLoaded` listener was ever registered to receive this event. Both tests still exercise the real toggle behavior, but the `if ( document.readyState === 'loading' )` branch (lines 67-70) has zero coverage.

To exercise it: `Object.defineProperty( document, 'readyState', { value: 'loading', configurable: true } )` before `require()`, then dispatch `DOMContentLoaded` and assert init ran via the listener path.

tests/js/wizard-lizard.test.js
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	beforeEach( () => {
+		document.body.innerHTML = '';
+		window.sessionStorage.clear();
```
kraftbj, line 22: `readStoredTheme` and `storeTheme` both wrap their `sessionStorage` calls in `try/catch` to handle the `SecurityError` that fires in restricted browser contexts (private mode, storage disabled, sandboxed iframes). The production comment explicitly calls this out as a real scenario, but neither catch path is exercised — so the safety net is currently load-bearing without any test pin.

A small spy that makes `window.sessionStorage.getItem` / `setItem` throw would lock both fallbacks in: `getItem` throwing should leave the wizard unthemed without propagating the exception; `setItem` throwing on click should still apply the in-memory class change.

tests/php/Admin/Onboarding_WizardTest.php
```diff
@@ -544,6 +544,8 @@ public function test_render_default_step_shows_destination_cards(): void {
 		$this->assertStringContainsString( 'Simple setup', $output );
 		$this->assertStringContainsString( 'Let people follow your site from apps like Mastodon. You can connect Bluesky later.', $output );
 		$this->assertStringContainsString( 'name="fosse_onboarding_destination"', $output );
+		$this->assertStringContainsString( 'data-fosse-lizard-toggle', $output );
+		$this->assertStringContainsString( '&#x1F98E;', $output );
```
kraftbj, line 548: While you're asserting toggle markup, consider also asserting the initial `aria-pressed="false"` and `aria-label="Onboarding Lizard"` (escaped via `esc_attr_e`). The aria-pressed initial value is load-bearing for the JS toggle contract — it's the value the script reads at startup before deciding whether to apply the theme. Currently no test pins either attribute, so a refactor that drops or alters them wouldn't fail anywhere.

src/Admin/assets/js/wizard-lizard.js
```diff
@@ -0,0 +1,74 @@
...
+			return;
+		}
+
+		const toggle = wizard.querySelector( TOGGLE_SELECTOR );
+
+		if ( ! toggle ) {
+			return;
+		}
+
+		if ( readStoredTheme() ) {
```
kraftbj, line 55: For users who enabled the egg earlier in the wizard flow, every step transition will briefly paint the unthemed wizard before this `applyTheme` runs (the JS is `in_footer`, so it executes after the wizard CSS has already painted). aria state stays in lockstep with the visual class so there's no a11y mismatch — this is purely cosmetic.

Acceptable to leave as-is for an easter egg; flagging only in case you ever want a no-flash experience (would mean inlining a tiny pre-paint script in the head, or rendering the initial themed state server-side via a companion cookie).

kraftbj (approved):
Approving — five small suggestions inline, all polish rather than blockers. Fine to land and follow up.

The change is well-scoped and aligns with the conventions established in recent wizard work:

- Wizard JS gated on the `admin_page_fosse-wizard` hook suffix, vanilla IIFE matching `wizard-appearance.js`
- `aria-pressed` toggle pattern from the PR 110 a11y pass
- `.fosse-wizard__lizard` added to the existing `prefers-reduced-motion: reduce` block
- sessionStorage-only state — no new option keys to teardown in E2E

A couple of smaller polish notes that probably aren't worth a follow-up on their own, just flagging in case the easter egg ever gets revisited:

- `.is-lizard-themed .button-primary` hover/focus rules hardcode hex values (`#315d12`, `#335d18`, `#264711`) instead of consuming `--fosse-wizard-accent` like the rest of the variant. The other 14 token overrides are clean — this one rule bypasses the pattern.
- The same codepoint (U+1F98E) appears in three encoding syntaxes across PHP, CSS, and the JS test fixture (`&#x1F98E;` vs `\1F98E`). `grep -r 1F98E` won't find the HTML-entity forms — a header comment on the lizard CSS section listing the codepoints would help future renames.

src/Admin/class-onboarding-wizard.php
RCowles: Addressed in #118 by pinning the hidden wizard style-toggle contract on the AP-unavailable render path.

tests/js/wizard-lizard.test.js
RCowles: Addressed in #118 by adding coverage for the delayed `DOMContentLoaded` initialization path.

tests/js/wizard-lizard.test.js
RCowles: Addressed in #118 by adding storage read/write failure coverage while keeping the in-memory UI state change intact.

tests/php/Admin/Onboarding_WizardTest.php
RCowles: Addressed in #118 by pinning the initial `aria-pressed` state and the generic accessible label for the hidden style toggle.

src/Admin/assets/js/wizard-lizard.js
RCowles: Leaving this as-is per your note. #118 keeps runtime behavior unchanged and only tightens coverage/maintainability.

RCowles:
Follow-up PR #118 covers the remaining non-blocking polish from the approval note: CSS variables for themed button states, a codepoint maintainability comment, storage-fallback coverage, delayed-init coverage, and render/ARIA assertions. The no-flash behavior is intentionally unchanged.

PR #118, opened 2026-05-09
Polish onboarding wizard style-toggle coverage

RCowles:
## Summary
- Add focused JS coverage for delayed initialization and storage-disabled fallback behavior.
- Pin the hidden wizard style-toggle render contract, including initial ARIA state, in PHP tests.
- Normalize themed primary-button colors through CSS variables and keep internal comments generic.

## Testing
- git diff --check
- pnpm run format:check
- pnpm run lint
- pnpm test

### DIFF
--- a/src/Admin/assets/css/admin.css
+++ b/src/Admin/assets/css/admin.css
@@ -1060,9 +1060,9 @@
 	--fosse-wizard-border-soft: #d9e8b8;
 	--fosse-wizard-border-faint: #e8f1cf;
 	--fosse-wizard-accent: #315d12;
-	--fosse-wizard-accent-border: #315717;
-	--fosse-wizard-accent-hover: #335d18;
-	--fosse-wizard-accent-hover-border: #264711;
+	--fosse-wizard-accent-button-border: #315717;
+	--fosse-wizard-accent-button-hover: #335d18;
+	--fosse-wizard-accent-button-hover-border: #264711;
 	--fosse-wizard-accent-soft: #eef7d6;
 	--fosse-wizard-accent-icon: #d9edab;
 	--fosse-wizard-accent-soft-border: rgba(49, 93, 18, 0.5);
@@ -1174,13 +1174,13 @@
 
 .fosse-wizard.is-lizard-themed .button-primary {
 	background: var(--fosse-wizard-accent);
-	border-color: var(--fosse-wizard-accent-border);
+	border-color: var(--fosse-wizard-accent-button-border);
 }
 
 .fosse-wizard.is-lizard-themed .button-primary:hover,
 .fosse-wizard.is-lizard-themed .button-primary:focus {
-	background: var(--fosse-wizard-accent-hover);
-	border-color: var(--fosse-wizard-accent-hover-border);
+	background: var(--fosse-wizard-accent-button-hover);
+	border-color: var(--fosse-wizard-accent-button-hover-border);
 }
 
 .fosse-wizard.is-lizard-themed .fosse-wizard__lizard {
--- a/src/Admin/assets/css/admin.css
+++ b/src/Admin/assets/css/admin.css
@@ -1051,7 +1051,7 @@
 	font-weight: 500;
 }
 
-/* Onboarding Lizard easter egg theme */
+/* Hidden wizard style-toggle theme */
 .fosse-wizard.is-lizard-themed {
 	--fosse-wizard-surface: #fbfff1;
 	--fosse-wizard-surface-muted: #f1f8df;
@@ -1060,6 +1060,9 @@
 	--fosse-wizard-border-soft: #d9e8b8;
 	--fosse-wizard-border-faint: #e8f1cf;
 	--fosse-wizard-accent: #315d12;
+	--fosse-wizard-accent-border: #315717;
+	--fosse-wizard-accent-hover: #335d18;
+	--fosse-wizard-accent-hover-border: #264711;
 	--fosse-wizard-accent-soft: #eef7d6;
 	--fosse-wizard-accent-icon: #d9edab;
 	--fosse-wizard-accent-soft-border: rgba(49, 93, 18, 0.5);
@@ -1118,6 +1121,7 @@
 	transform: translateY(-1px);
 }
 
+/* Theme glyphs: U+1F95A egg, U+1F98E reptile, U+1F43E paw. */
 .fosse-wizard.is-lizard-themed .fosse-wizard__progress-dot::before {
 	content: "\1F95A";
 	width: auto;
@@ -1170,13 +1174,13 @@
 
 .fosse-wizard.is-lizard-themed .button-primary {
 	background: var(--fosse-wizard-accent);
-	border-color: #315717;
+	border-color: var(--fosse-wizard-accent-border);
 }
 
 .fosse-wizard.is-lizard-themed .button-primary:hover,
 .fosse-wizard.is-lizard-themed .button-primary:focus {
-	background: #335d18;
-	border-color: #264711;
+	background: var(--fosse-wizard-accent-hover);
+	border-color: var(--fosse-wizard-accent-hover-border);
 }
 
 .fosse-wizard.is-lizard-themed .fosse-wizard__lizard {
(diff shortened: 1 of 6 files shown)


tests/js/wizard-lizard.test.js
```diff
@@ -9,52 +11,112 @@ describe( 'Wizard lizard easter egg', () => {
...
 		renderWizard();
 		loadScript();
 
-		const wizard = document.querySelector( '.fosse-wizard' );
-		const toggle = document.querySelector( '[data-fosse-lizard-toggle]' );
+		const { wizard, toggle } = getElements();
+
+		expect( wizard.classList.contains( 'is-lizard-themed' ) ).toBe( false );
+		expect( toggle.getAttribute( 'aria-pressed' ) ).toBe( 'false' );
+	} );
```
kraftbj: This test reads as coverage for the storage-read-failure path, but both assertions pass trivially: the fixture in `renderWizard()` hardcodes `aria-pressed="false"` and the class is never added on init, so the test still passes if `init()` had silently early-returned and never attached a click listener at all.

Suggest restoring the spy and clicking the toggle to confirm the listener actually wired up despite the read failure:

```suggestion
	test( 'ignores storage read failures and still attaches the click handler', () => {
		const getItemSpy = jest
			.spyOn( Storage.prototype, 'getItem' )
			.mockImplementation( () => {
				throw new Error( 'storage unavailable' );
			} );

		renderWizard();
		loadScript();

		const { wizard, toggle } = getElements();

		expect( wizard.classList.contains( 'is-lizard-themed' ) ).toBe( false );
		expect( toggle.getAttribute( 'aria-pressed' ) ).toBe( 'false' );

		getItemSpy.mockRestore();
		toggle.click();

		expect( wizard.classList.contains( 'is-lizard-themed' ) ).toBe( true );
		expect( toggle.getAttribute( 'aria-pressed' ) ).toBe( 'true' );
	} );
```

src/Admin/assets/css/admin.css
```diff
@@ -1060,6 +1060,9 @@
 	--fosse-wizard-border-soft: #d9e8b8;
 	--fosse-wizard-border-faint: #e8f1cf;
 	--fosse-wizard-accent: #315d12;
+	--fosse-wizard-accent-border: #315717;
+	--fosse-wizard-accent-hover: #335d18;
+	--fosse-wizard-accent-hover-border: #264711;
```
kraftbj: The surrounding accent palette uses an adjective-qualifier pattern (`--fosse-wizard-accent-soft`, `-accent-icon`, `-accent-soft-border`). The three new tokens introduce a different shape (`-accent-border`, `-accent-hover`, `-accent-hover-border`) that conflates "accent" with the button-specific use and makes the next contributor pick between two conventions.

Suggest scoping the names to `-accent-button-…` so the scope is explicit and the qualifier order stays consistent:

```suggestion
	--fosse-wizard-accent-button-border: #315717;
	--fosse-wizard-accent-button-hover: #335d18;
	--fosse-wizard-accent-button-hover-border: #264711;
```

The two `.button-primary` usages further down (around lines 1177 and 1182-1183) would need the same rename — happy to ship that as a follow-up suggestion once you decide on naming. Non-blocking; flagging while the convention is fresh.

tests/js/wizard-lizard.test.js
```diff
@@ -9,52 +11,112 @@ describe( 'Wizard lizard easter egg', () => {
...
+		const { wizard, toggle } = getElements();
+
+		expect( wizard.classList.contains( 'is-lizard-themed' ) ).toBe( false );
+
+		document.dispatchEvent( new Event( 'DOMContentLoaded' ) );
+		toggle.click();
+
+		expect( wizard.classList.contains( 'is-lizard-themed' ) ).toBe( true );
+		expect( toggle.getAttribute( 'aria-pressed' ) ).toBe( 'true' );
+	} );
```
kraftbj, line 80: Heads up: the `DOMContentLoaded` listener registered by the IIFE during this test stays attached to `document` after the test ends — no `{ once: true }` in `wizard-lizard.js`, no `removeEventListener`, and `beforeEach` doesn't flush document listeners. Harmless today, but the moment any future test re-dispatches `DOMContentLoaded`, you'll get two `init()` calls (one from this test's stale closure against an empty DOM, one from the new test) and the assertions will get weird in a hard-to-debug way.

Two clean ways to fix:

1. Switch the listener in `src/Admin/assets/js/wizard-lizard.js` to `document.addEventListener( 'DOMContentLoaded', …, { once: true } )`. Production-side change but also semantically correct (the listener has no reason to fire twice).
2. Keep production code as-is and restructure this test to call `loadScript( 'loading' )` before `renderWizard()`, assert the toggle is inert pre-event, then dispatch `DOMContentLoaded` and assert it becomes functional — that way the listener fires exactly once during the test's intended window.

P3 / non-blocking. Calling it out now because the next person to add a `'loading'` test will trip on it.

tests/js/wizard-lizard.test.js
```diff
@@ -9,52 +11,112 @@ describe( 'Wizard lizard easter egg', () => {
...
+			value: readyState,
+			configurable: true,
+		} );
+
 		jest.isolateModules( () => {
 			require( '../../src/Admin/assets/js/wizard-lizard.js' );
 		} );
-
-		document.dispatchEvent( new Event( 'DOMContentLoaded' ) );
 	}
```
kraftbj, line 29: `Object.defineProperty( document, 'readyState', … )` mutates the descriptor but isn't restored by `jest.restoreAllMocks()` in `beforeEach` — only `jest.spyOn` mocks are. The reason it works today is that every test calls `loadScript()` which unconditionally overwrites the value, so the leak is hidden. The juxtaposition reads as if cleanup is happening when it isn't.

Jest 30 (already in `package.json`) ships `jest.replaceProperty`, which IS reset by `restoreAllMocks`:

```suggestion
	function loadScript( readyState = 'complete' ) {
		jest.replaceProperty( document, 'readyState', readyState );

		jest.isolateModules( () => {
			require( '../../src/Admin/assets/js/wizard-lizard.js' );
		} );
	}
```

kraftbj (commented):
Polish PR is in good shape overall — most of the contract-pinning lands well, and the storage-failure tests pull useful coverage. Two P2 nits and two P3 follow-ups inline; only the storage-read-failure test (the first comment) is the one I'd actually want addressed before merge — it currently passes whether or not `init()` runs at all. The rest are happy-to-defer.

For what it's worth, the deliberate decision to keep `STORAGE_KEY`, the `is-lizard-themed` class, `data-fosse-lizard-toggle`, and the filename unchanged is the right call — preserving stored-session compatibility is worth the surface vocabulary drift, and the docblock/method-name mismatch on `render_lizard_toggle()` is fine to leave alone.

tests/js/wizard-lizard.test.js
RCowles: Addressed in 3fac5b3. I expanded this test so the storage-read failure path now restores the `getItem` spy, clicks the toggle, and asserts the theme class plus `aria-pressed` flip to prove the click handler was attached despite the read failure.

src/Admin/assets/css/admin.css
RCowles: Addressed in 3fac5b3. I renamed the themed button vars to the `--fosse-wizard-accent-button-*` shape and updated the `.button-primary` background/border usages to match, so the token scope is explicit and the old `accent-border` / `accent-hover` names are gone.

tests/js/wizard-lizard.test.js
RCowles: Addressed in 3fac5b3. I took the production-side option and added `{ once: true }` to the `DOMContentLoaded` listener in `wizard-lizard.js`, so the early-load listener cannot linger and re-run on later dispatched events.

tests/js/wizard-lizard.test.js
RCowles: Addressed in 3fac5b3 with the same cleanup intent, but using `jest.spyOn( document, "readyState", "get" ).mockReturnValue( ... )` instead of `jest.replaceProperty()`. In jsdom, `document.readyState` is exposed as a getter and Jest rejects `replaceProperty()` for it; the getter spy is restored by `jest.restoreAllMocks()`.

